PR #2, opened 2014-05-19
Add sign-up cancellation configuration

eugeneventer:
Add ability to disable sign-up cancellations for a face-to-face
activity.

Add capability to be assigned to roles who should be able to configure
cancellation (at activity and session level) availability.


danmarsden:
I rebased this and did some initial testing - seemed to work ok - merged into master branch. Thanks.

PR #11, opened 2015-05-04
Renamed class name from signup to signup_success in order to fix blank s...

mvazquezc:
...creen signup.php error caused by class cannot being created because name mismatch


danmarsden:
this has already been merged in, thanks! - closing.

PR #13, opened 2015-09-17
Fixed name of signup_success class

andrewhancox:
There's currently a bug which causes an error when signing up. The class signup_success is referred to in the code and a class file exists called signup_success but the class within it is signup.


PeloNZ:
Thanks Andrew

PR #14, opened 2016-02-03
Added Travis CI

Mihailoff:
Please check out the results of code syntax and other checks. See Moodle forum thread https://moodle.org/mod/forum/discuss.php?d=323384


danmarsden:
thanks - we've added a travis file to the master branch for now - expecting it to throw a lot of errors but something to work on in future.

PR #16, opened 2017-06-05
Update session_form to remove undefined calls

fsjoyti:


danmarsden:
Thanks heaps for taking a look at this but it doesn't look complete to me - the original form is supposed to allow multiple session times to be added and then it iterates over each one to check if they are valid. if they are not valid it adds an item to the errors array for that item. which won't work with this patch as it doesn't appear complete. I also suspect mulitple session times being added isn't working correcltly either so we probably need to look at that as a complete issue. Closing this pull request as it is not complete.

PR #22, opened 2017-10-02
Add search functionality

sk-unikent:
Not sure if this is being intentionally left out?


danmarsden:
not intentional - thanks for the pull request, master branch supports 3.2 and higher and you have used the base search class for 3.2 and higher so I've merged that in to master only. Thanks again!

PR #23, opened 2017-10-02
Remove deprecated method call

sk-unikent:
Since 3.0, so won't change STABLE branches, master already requires 3.2+


danmarsden:
cool - thanks for the pull request!

PR #24, opened 2018-01-17
mod/facetoface: Replace disabledif with hideif

justine-lr:
Have replaced all instances of disabledif with hideif. 
Have **NOT TESTED** this with ANY versions.

### DIFF
--- a/classes/session_form.php
+++ b/classes/session_form.php
@@ -98,8 +98,8 @@ class mod_facetoface_session_form extends moodleform {
         $repeatcount = $this->_customdata['nbdays'];
 
         $repeatoptions = array();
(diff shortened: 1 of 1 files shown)


danmarsden:
Hi @justine-lr - thanks for the pull request.
I've just tested this and unfortunately it doesn't work - that particular disabledif call is slightly different - the other academy students were replacing disabledif calls like this one:
https://github.com/catalyst/moodle-mod_facetoface/blob/dedebc41a9f08f31b811a8922d89568ad4164103/mod_form.php#L107

If you would like to spend some further time on this feel free to file a new pull request changing those ones to hideif (we love getting pull requests!) and I'll take another look. I hope you have found the academy interesting so far!

PR #28, opened 2018-06-04
Fixed calendar events, now compatible with Moodle 3.2 to 3.5.

michael-milette:
3.2.2 (2018-05-30):
Bug fixes:
- Site, course and user calendar events now work properly. They no longer all end up being course events.
- Creating, updating and deleting Face-to-face sessions no longer leave orphan and duplicate events in the events table.
- Deleting a Face-to-face activity from a course no longer leave orphan calendar events in the events table.

Improvements:
- Now compatible with Moodle 3.2 to 3.5

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,3 +1,12 @@
+3.2.2 (2018-05-30):
+Bug fixes:
+- Site, course and user calendar events now work properly. They no longer all end up being course events.
+- Creating, updating and deleting Face-to-face sessions no longer leave orphan and duplicate events in the events table.
+- Deleting an instance of Face-to-face from a course no longer leave orphan calendar events in the events table.
+
+Improvements:
+- Now compatible with Moodle 3.2 to 3.5
+
 2.3.1 (2014-03-06):
 Bug fixes:
 - fix SQL group by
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks! - merging it in now. code looks like it makes sense - haven't done a lot of testing on it but I'm sure we'll get some feedback if anything comes up - is there any clean up that we should be doing on existing events in the calendar?

michael-milette:
Hi @danmarsden ,

Thanks for the merge! 👍 

I am fortunate that I am starting to implement it in my project so I was able to start with empty tables and not have to worry about too much cleanup - which I would do manually during debugging until I fixed the code.

Unfortunately anyone who has been using face-to-face for a while will have tons of entries in the database table as it was creating multiple entries for every session and almost never deleting them. There was also lots of entries which were also being left behind when you would delete a session or the instance of face-to-face.

If you are planning to automate the cleanup process during an upgrade process, this would require some thought and planning. I think the easiest way would be to put every existing session in every instance of face-to-face in every course through the session update process, as if you had edited each session and, without making any changes, just saved it. That should reset all of the existing sessions (to be confirmed). The same could be accomplished if you manually edited a session and then saved it.

You should be cautioned though that Moodle administrators might be surprised at how many events suddenly show up in the Events calendar after the cleanup. Example, in instances where they have chosen in the settings to add events to the site calendar, they will find many new events there. Anyone enrolled in the course will also see a second entry in the events calendar. I wish there was a way to have only one event show up in the calendar instead of two, however, that is a limitation of the Moodle calendar.

It might be wise to warn people about what will happen when they upgrade.

As for orphan sessions left behind from deleted sessions or deleted instances of face-to-face, that would be a different story, but I think not impossible.

Now this is just off the top of my head but, I thin that if you simply delete all of the face-to-face events in the events calendar and then re-create them all based on the information available in the face-to-face tables (it should be all there), that might just take care of both repair and cleanup. Alternatively, you could scan through the events table and selectively delete those face-to-face events that are not in any of the face-to-face tables.

Just some food for thoughts.

Thanks again! :-)

Michael

PR #33, opened 2018-07-23
Date fix

### LINKED ISSUE #32: date export issues
After latest changes there is an issue with session dates when exporting to excel :

( ! ) Notice: A non well formed numeric value encountered in /var/www/xxx/lib/moodlelib.php on line 2275
--

1 | 0.0003 | 361336 | {main}( ) | .../view.php:0
2 | 0.0383 | 6013168 | facetoface_download_attendance( ) | .../view.php:80
3 | 0.0519 | 7162496 | facetoface_write_activity_attendance( ) | .../lib.php:1294
4 | 0.0618 | 7350560 | facetoface_write_activity_attendance_helper( ) | .../lib.php:1560
5 | 0.0642 | 7361520 | MoodleExcelWorksheet->write_date( ) | .../lib.php:1632
6 | 0.0642 | 7361520 | usergetdate( ) | .../excellib.class.php:230
7 | 0.0642 | 7361520 | getdate ( )

Will submit fix soon

ilya-catalyst:
#32

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -1569,7 +1569,13 @@ function facetoface_write_activity_attendance_helper(&$worksheet, $i, $session, 
     $finishtime  = get_string('wait-listed', 'facetoface');
 
     if ($session->datetimeknown) {
-        $sessiondate = userdate($session->timestart, get_string('strftimedate', 'langconfig'));
+        if (method_exists($worksheet, 'write_date')) {
+            // Needs the patch in MDL-20781.
+            $sessiondate = (int)$session->timestart;
+        } else {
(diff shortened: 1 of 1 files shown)


danmarsden:
hmm - I saw that reference to the bug tracker issue in the comments which has already been fixed which confused me and the code looks quite strange - I'm not sure we should replace all that code again - we should probably rewrite it a bit. do you have time to tidy that up a bit better?

danmarsden:
I took another look and saw what I'd done wrong... and tested it this time! :-) - have pushed a fix thanks.

PR #36, opened 2018-11-05
Update attendees.php

adigitalife:
Sort attendees alphabetically according to firstname.


danmarsden:
Thanks for the pull request. Unfortunately to change the sorting of the page we'd really need to implement a user-sort function - allowing the user to click on the column headers to change the sorting. Preferably by using the core Moodle apis that allow for similar behaviour.

PR #56, opened 2021-04-27
Add rich text editor for confirmation messages

### LINKED ISSUE #55: Customizing the F2F notification messages to allow rich text emails
Hello,

Currently the F2F notifications coming out of the plugin are plain text, however we would like to customize this to include media such as images and colors. An idea is by incorporating the Moodle Atto editor that allows rich text to be created. 

We're planning to begin by modifying the plaintext field in the session confirmation email section to instead used the Moodle Atto editor. Following which we plan to modify any of its other dependencies to accept the rich text and send out better looking emails.

This is an issue we can attempt to work on with a guidance and review from your team. 😄

chalhoub-university:
Fix: #55 

Hello! 
We have replaced the plain 'text area' with a rich text editor that allows you to add html formatted messages for the confirmation.
Currently, HTML Messages can be created & updated for the confirmation message. These messages are also sent out as rich text to the users. 

We would like a review on the above and assistance with one area we couldn't figure out: 

1. With using the rich text editor, we were not able to set a default value for the editor
2. Retrieve an existing value from the DB to display in the editor when a user tries to update the confirmation message at a later stage.

If there are any suggestions on accomplishing the above, we would love to work to fix those further and replicate it for the other types of messages (waitlist, cancellation).

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -46,8 +46,8 @@ function facetoface_send_admin_upgrade_msg($data) {
 
     $table = new html_table();
     $table->head = array('Custom field ID',
-                         'Custom field original shortname',
-                         'Custom field new shortname');
+        'Custom field original shortname',
+        'Custom field new shortname');
     $table->data = $data;
     $table->align = array ('center', 'center', 'center');
 
@@ -745,6 +745,21 @@ function xmldb_facetoface_upgrade($oldversion=0) {
 
         // Facetoface savepoint reached.
         upgrade_mod_savepoint(true, 2017053000, 'facetoface');
+    }
+
+    if ($oldversion < 2020080401) {
+
+        // Define field confirmationmessageformat to be added to facetoface.
+        $table = new xmldb_table('facetoface');
+        $field = new xmldb_field('confirmationmessageformat', XMLDB_TYPE_INTEGER, '2', null, XMLDB_NOTNULL, null, '0', 'confirmationmessage');
+
+        // Conditionally launch add field confirmationmessageformat.
+        if (!$dbman->field_exists($table, $field)) {
+            $dbman->add_field($table, $field);
+        }
+
+        // Facetoface savepoint reached.
+        upgrade_mod_savepoint(true, 2020080401, 'facetoface');
     }
 
     return $result;
--- a/lang/en/facetoface.php
+++ b/lang/en/facetoface.php
@@ -410,6 +410,7 @@ To re-schedule your booking you need to cancel this booking and then re-book a n
 
 You will receive a reminder [reminderperiod] business days before this course.
 ';
+$string['setting:defaultconfirmationmessagedefault2'] = '<p><b>Thank you for signing up!</b>';
 $string['setting:defaultconfirmationsubject'] = 'Default subject line for confirmation emails.';
 $string['setting:defaultconfirmationsubject_caption'] = 'Confirmation subject';
 $string['setting:defaultconfirmationsubjectdefault'] = 'Course booking confirmation: [facetofacename], [starttime]-[finishtime], [sessiondate]';
--- a/lib.php
+++ b/lib.php
@@ -314,6 +314,7 @@ function facetoface_add_instance($facetoface) {
     global $DB;
 
     $facetoface->timemodified = time();
+    $facetoface->confirmationmessage = $facetoface->confirmationmessage['text'];
     facetoface_fix_settings($facetoface);
     if ($facetoface->id = $DB->insert_record('facetoface', $facetoface)) {
         facetoface_grade_item_update($facetoface);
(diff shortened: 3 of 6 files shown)


marcghaly:
Hi @chalhoub-university @geoffreyvanwyk, I m currently working on this issue, could you please provide me with more details about point `2`. From which table would you like to get data in the DB ?

Regards,

db/upgrade.php
```diff
@@ -46,8 +46,8 @@ function facetoface_send_admin_upgrade_msg($data) {
 
     $table = new html_table();
     $table->head = array('Custom field ID',
-                         'Custom field original shortname',
-                         'Custom field new shortname');
+        'Custom field original shortname',
```
mattporritt, line 49: I know this isn't really a change in this tracker. I only noticed because the indentation has been changed. However, these should be lang strings instead of hard coded.
A good idea here would be to open a new issue to fix this.

lang/en/facetoface.php
```diff
@@ -410,6 +410,7 @@
 
 You will receive a reminder [reminderperiod] business days before this course.
 ';
+$string['setting:defaultconfirmationmessagedefault2'] = '<p><b>Thank you for signing up!</b>';
```
mattporritt, line 413: There is an unclosed <p> tag here, that will need to be fixed.
Ideally, there wouldn't be HTML in lang strings. But sometimes it is unavoidable.

lib.php
```diff
@@ -1984,7 +1986,7 @@ function facetoface_user_cancel($session, $userid=false, $forcecancel=false, &$e
  * @returns string Error message (or empty string if successful)
  */
 function facetoface_send_notice($postsubject, $posttext, $posttextmgrheading,
-                                $notificationtype, $facetoface, $session, $userid) {
+                                $notificationtype, $facetoface, $session, $userid,$htmlmessage) {
```
mattporritt: Need to update the PHP docs for this function to include the new variable $htmlmessage.
Also `$userid,$htmlmessage` needs to be this with a space `$userid, $htmlmessage`

lib.php
```diff
@@ -2035,7 +2037,8 @@ function facetoface_send_notice($postsubject, $posttext, $posttextmgrheading,
                                                           $user, $session, $session->id);
                 $body = facetoface_email_substitutions($posttext, format_string($facetoface->name), $facetoface->reminderperiod,
                                                        $user, $session, $session->id);
-                $htmlbody = ''; // TODO.
+                 $htmlmessage = facetoface_email_substitutions($posttext, $facetoface->name, $facetoface->reminderperiod, $user, $session, $session->id);
```
mattporritt: fix the indentation here

lib.php
```diff
@@ -2048,7 +2051,8 @@ function facetoface_send_notice($postsubject, $posttext, $posttextmgrheading,
                                                       $user, $session, $session->id);
             $body = facetoface_email_substitutions($posttext, format_string($facetoface->name), $facetoface->reminderperiod,
                                                    $user, $session, $session->id);
-            $htmlbody = ''; // FIXME.
+             $htmlmessage = facetoface_email_substitutions($posttext, $facetoface->name, $facetoface->reminderperiod, $user, $session, $session->id);
```
mattporritt: fix the indentation here

lib.php
```diff
@@ -2154,8 +2158,11 @@ function facetoface_send_confirmation_notice($facetoface, $session, $userid, $no
     // Set invite bit.
     $notificationtype |= MDL_F2F_INVITE;
 
+    //Set HTML Body
```
mattporritt: wrong comment doc style

mattporritt (changes requested):
The logic here is largely sound.
The main things I've highlighted are code style and PHP docs. The automated tests in the Github actions should also be picking these up.
This plugin has existing unit and Behat tests. It would be good to extent at least the unit tests to cover the new functionality.

db/upgrade.php
marcghaly: Hi @mattporritt, 
This is fixed in another issue #74.

lang/en/facetoface.php
marcghaly: This is also fixed in issue #74

lib.php
marcghaly: Fixed.

lib.php
marcghaly: Fixed

lib.php
marcghaly: Fixed

lib.php
marcghaly: Fixed.

marcghaly:
Hi @mattporritt,

CS issues have been fixes, and I have created another issue and PR for lang strings #74.

Cheers,

Issue #35, opened 2018-08-16
Export Attendance not population all columns


nsheehan:
While testing face-to-face we have added Custom Session Fields. They are Location, Venue and Room. The check box to show in exports and list is selected however when we export while the columns are created the values are blank.
Is this expected behaviour?
![missing-f2f-data](https://user-images.githubusercontent.com/8668047/44220221-fc298a00-a14b-11e8-9b85-4d8d1e786034.png)
![f2f-location-venue](https://user-images.githubusercontent.com/8668047/44220234-0055a780-a14c-11e8-9a3c-e104a8fceb11.png)

danmarsden:
Thanks for the report - sounds like a bug! - if you figure out the cause feel free to submit a pull request with a fix! - unfortunately it's unlikely we'll be able to spend time on a fix for this unless someone funds the time required for one of our developers to investigate. Please get in touch if you would like to fund any development time on the face-to-face activity.

danmarsden:
just adding a link to your forum discussion as for future reference:
https://moodle.org/mod/forum/discuss.php?d=374719

PR #58, opened 2021-07-08
Export Attendance Custom Session Fields

sameer-ah:
Fix: #35
-Custom session fields values are populated in export attendance columns.


danmarsden:
Thanks heaps for the PR - looks good to me - merged!

PR #61, opened 2021-08-11
Custom Session Fields Language Filter Fix

sameer-ah:
Hello,

There are some minor fixes for custom session fields language filter. _filter_string_ is applied to _$field->name_.

Please let me know in case of any questions.

Thanks,
Sameer


danmarsden:
thanks for the PR - makes sense to me - merged it in.

Issue #63, opened 2021-10-07
get_all_user_name_fields() has been deprecated in Moodle 3.11


michael-milette:
Adding a new session using FaceToFace on 3.11 throws the following errors:

get_all_user_name_fields() is deprecated. Please use the \core_user\fields API instead
line 3384 of /lib/deprecatedlib.php: call to debugging()
line 3722 of /mod/facetoface/lib.php: call to get_all_user_name_fields()
line 159 of /mod/facetoface/classes/session_form.php: call to facetoface_get_trainers()
line 214 of /lib/formslib.php: call to mod_facetoface_session_form->definition()
line 143 of /mod/facetoface/sessions.php: call to moodleform->__construct()

get_all_user_name_fields() is deprecated. Please use the \core_user\fields API instead
line 3384 of /lib/deprecatedlib.php: call to debugging()
line 167 of /mod/facetoface/classes/session_form.php: call to get_all_user_name_fields()
line 214 of /lib/formslib.php: call to mod_facetoface_session_form->definition()
line 143 of /mod/facetoface/sessions.php: call to moodleform->__construct()

See https://tracker.moodle.org/browse/MDL-45242 for deprecation notice:

The corresponding existing functions are all deprecated: get_extra_user_fields(), get_extra_user_fields_sql(), get_user_field_name(), get_all_user_name_fields(), and user_picture::fields(). All the features from these functions are subsumed into the new API:

https://docs.moodle.org/dev/User_fields

Best regards,

Michael

danmarsden:
Thanks Michael - PR's welcome :-) - we haven't got anyone here using this in 3.11 so it's unlikely we'd get to it ourselves - thanks!

michael-milette:
Thank you!

PR #64, opened 2021-11-07
Fix-63: Fix deprecated get_all_user_name_fields() in Moodle 3.11.

michael-milette:
Hi @danmarsden ,

This pull request addresses the issue described in #63 relating to the function get_all_user_name_fields() which is deprecated as of Moodle 3.11.

Note: The fix for this plugin continues to maintain backwards compatibility with previous versions of Moodle.

Please let me know if you have any questions or concerns.

Best regards,

Michael

### DIFF
--- a/classes/session_form.php
+++ b/classes/session_form.php
@@ -164,7 +164,7 @@ class mod_facetoface_session_form extends moodleform {
                 $rolename = $rolename->name;
 
                 // Attempt to load users with this role in this course.
-                $usernamefields = get_all_user_name_fields(true);
+                $usernamefields = facetoface_get_all_user_name_fields(true);
                 $rs = $DB->get_recordset_sql("
                     SELECT
                         u.id,
--- a/editattendees.php
+++ b/editattendees.php
@@ -100,7 +100,7 @@ if (optional_param('add', false, PARAM_BOOL) && confirm_sesskey()) {
                 }
             }
 
-            $usernamefields = get_all_user_name_fields(true);
+            $usernamefields = facetoface_get_all_user_name_fields(true);
             if (facetoface_get_user_submissions($facetoface->id, $adduser)) {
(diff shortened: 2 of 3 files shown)


danmarsden:
thanks @michael-milette !! - For catalyst managed plugins we prefer to create a new branch within the repository rather than introduce polyfill style handling as this creates technical debt that needs cleaning up in future. If you get a chance to reformat that to just call the new 3.11 functions directly that would be great - I can then create an older branch in github for the older versions of Moodle to use.
No worries if you don't get a chance to do this, we'll probably need to get to it at some point...

michael-milette:
Hi @danmarsden ,

Done. I removed the 3 extra lines of code that provided backwards compatibility.

Best regards,

Michael

danmarsden:
thanks Michael - I'm not a big fan of wrapper functions like that and I usually prefer to just call API in place, but... it does fix it so I might end up merging it as is - few other things to clear this week but will try to come back to it later.

mattporritt:
Hi @danmarsden ,
I've had a quick look at this, the proposed solution is OK. The new API doesn't have a direct replacement to get the sql used here. It does have a `get_sql` method, but it's not the same.  The wrapper function here does save some code duplication without the need to do a big refactor.

I've made a MOODLE_311_STABLE branch in this repo and have cherry picked in @michael-milette change.

I've also added github actions to this 3.11 branch and updated the branch readme.

The github actions are currently failing, I'm going to get a member of the Cat CA team to look at this, just to get things in a better shape.  (Why not or behat tests?)

Cheers,
Matt P

danmarsden:
closing this as it's been merged into that new branch manually - thanks @mattporritt & @michael-milette

michael-milette:
Thank you @danmarsden and @mattporritt !

PR #66, opened 2021-11-15
Language Filter Fix

sameer-ah:
Hello, 

Some language filter fixes at
-Locations dropdown list (shown if there are more than 2 locations)
-Session details text
-Calendar event title, description and location string
-Custom fields dropdown list values
-Reminder message title

_filter_string_ is applied to fix the issues. 

Thanks,
Sameer

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -780,7 +780,7 @@ function facetoface_email_substitutions($msg, $facetofacename, $reminderperiod, 
     if (empty($data->details)) {
         $msg = str_replace(get_string('placeholder:details', 'facetoface'), '', $msg);
     } else {
-        $msg = str_replace(get_string('placeholder:details', 'facetoface'), html_to_text($data->details), $msg);
+        $msg = str_replace(get_string('placeholder:details', 'facetoface'), html_to_text(format_text($data->details)), $msg);
     }
     $msg = str_replace(get_string('placeholder:reminderperiod', 'facetoface'), $reminderperiod, $msg);
 
@@ -900,9 +900,9 @@ function facetoface_cron() {
             continue;
         }
 
-        $postsubject = facetoface_email_substitutions($postsubject, $signupdata->facetofacename, $signupdata->reminderperiod,
+        $postsubject = facetoface_email_substitutions($postsubject, format_string($signupdata->facetofacename), $signupdata->reminderperiod,
                                                       $user, $signupdata, $signupdata->sessionid);
-        $posttext = facetoface_email_substitutions($posttext, $signupdata->facetofacename, $signupdata->reminderperiod,
+        $posttext = facetoface_email_substitutions($posttext, format_string($signupdata->facetofacename), $signupdata->reminderperiod,
                                                    $user, $signupdata, $signupdata->sessionid);
         $posttextmgrheading = facetoface_email_substitutions($posttextmgrheading, $signupdata->facetofacename, $signupdata->reminderperiod,
                                                              $user, $signupdata, $signupdata->sessionid);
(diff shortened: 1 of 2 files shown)


danmarsden:
Thanks for the PR - your code doesn't look correct which makes me wonder if you've tested it properly? - I'm happy to merge something like this in, but in at least one place you have changed, format_string will not be doing anything.

sameer-ah:
All the code changes are tested in our test environment. Could you please point out which one format_string, will not do anything. All of the fixes were made against issues raised due to language tags and were fixed after applying this fix.

danmarsden:
I deliberately didn't point it out because you will just make that change and I don't really have time to test this patch myself - it should be pretty obvious to find if you get someone that understands Moodle functions to peer review it for you internally though - take a closer look and you'll probably find it yourself.

sameer-ah:
I have removed extra _format_string_ used at https://github.com/catalyst/moodle-mod_facetoface/blob/70d53184c75228c3896b440476ebf9ad966e99ce/lib.php#L907 Thanks for bringing it up.

danmarsden:
thanks - I didn't even get to that line when reviewing your Pull request so interesting to hear you had to remove that... which does still suggests (to me) that you haven't tested each change in your PR? :-)

take another look at the very first change in this PR - does multi-lang content actually work there? - I can't see how it would - look closely at what happens to the var in that line before you pass it to format_string

danmarsden:
and for a similar reason - I'm not sure this line will be working either:
https://github.com/catalyst/moodle-mod_facetoface/pull/66/files#diff-8f5ff72ef1cf59199cfc4bbf16f2475d09e448601242f0d2c51bb94bd4aa1395R2758
(the one above "might")

sameer-ah:
The reason I removed that line was because it was not needed by us but I made that change to be consistent with previous two lines. Upon further testing (as suggested by you), it turned out that with current default manager header text, _format_string_ is not doing anything. Although, it might be needed if someone changes text to include session name containing language filters. Based on your comments, I removed it to make sure there is no extra _format_string_. 

As for as using _format_string_ for _$data->details_ **var**, yes it is needed and works here, at least in our case. We are using [Multi-Language](https://moodle.org/plugins/filter_multilang2) plugin and formatting text with _format_string_ does translate text instead of showing **mlang** tags. It has been tested (by myself, second developer and testing team) and we did not find any issues, first passing **var** to _html_to_text_ function and then passing the result to _format_string_ to translate language tags. 

Kindly let me know in case I am still missing something to make sure this PR is not breaking existing code.

danmarsden:
Thanks Sameer - adding format_string should allow the standard Moodle filters to work - including the standard Multi-lang filter which uses html to separate and identify the different languages. Your code supplied here does not add support for standard core Multi-lang as you have the format_string call in the wrong place. If you test with the standard core multi-lang filter instead of your custom one you would probably see it failing - there are also several other filter types that would fail to work as well.

format_string should be called before any further string manipulation - pass the var to format string first - then pass it to html_to_text.

sameer-ah:
Thanks Dan. I really appreciate providing details how Moodle filters work. I tested code with standard Multi-lang filter and it is failing. Furthermore, changing the order of _format_string_ and _html_to_text_ is also not fixing the issue. However, replacing _format_string_ with _format_text_ where html text can be provided (in this case **var** $data->details or $session->details), is working for both standard as well as custom multi-lang filters. I have made these changes along with passing **var** to _format_text_ first before passing to _html_to_text_.

danmarsden:
yeah that looks more like I'd expected to see - nice work :-)

PR #77, opened 2022-03-18
Adding multi-search/delimited search

vedantnayar:
Using the | character, users can search the list of attendees for multiple users at the same time. Particularly useful when the list of users is huge, to avoid needing to enrol users one at a time.

Recommendation: Works well with the Excel TEXTJOIN() function

Contributions by @chalhoub-university, @geoffreyvanwyk and me

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -4177,7 +4177,7 @@ class facetoface_candidate_selector extends user_selector_base {
      * @param <type> $search
      * @return array
      */
-    public function find_users($search) {
+    public function find_users_base($search) {
         global $DB;
 
         // All non-signed up system user.
@@ -4187,7 +4187,7 @@ class facetoface_candidate_selector extends user_selector_base {
         $countfields = 'SELECT COUNT(u.id)';
         $sql = "
                   FROM {user} u
-                 WHERE $wherecondition
+                  WHERE $wherecondition
                    AND u.id NOT IN
                        (
                        SELECT u2.id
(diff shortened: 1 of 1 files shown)


danmarsden:
Thanks for the PR - my initial instinct (without looking closely) would be that this sounds wrong as I can't think of any other instance in Moodle core where a search is done in a similar way. If the page is hard to manage with multiple users, there are probably other ways to improve the UX without building in "hidden features" like this..

vedantnayar:
Hi Dan,
This works well for our team's use case, but I'd definitely like to hear of the right way to implement this so we can make improvements so that it's more useful to everyone. Cheers!

danmarsden:
from memory, I think this is using the same UI as the manage site roles page - it might be better to look at moving towards something more like the enrol users UI or somehow use some similar UX to the new moodle 4.0 reportbuilder API, but ideally we'd get Moodle to improve the UX of a core component and then copy that within F2F - 
Moodle 4.0 has some new reportbuilder apis with various pop-out filters - would be nice if you could apply various filters to the list of users being displayed, but it really depends on the use-case here..

Of course - if you manage to get something like "|" supported in the moodle core manage site roles page, we could then add the same behaviour into the F2F version of the plugin too.

danmarsden:
one option would be to add a new "search option" checkbox that enable multi-search for searcing with any individual word - eg if the checkbox was ticked a search of "dan amy" was entered it would find all users with "dan" and all users with "amy" 

Also - anything like this being added would likely need unit tests to test the new behaviour too.

danmarsden:
closing this one as won't fix for now - but feel free to resubmit if you are able to improve the implementation as mentioned above - thanks!

PR #82, opened 2022-06-23
Fix multiple CSS issues reported in issue 81

skodak:


danmarsden:
Thanks Petr! Fyi we have hardly anyone at Catalyst using f2f in moodle - workplace has a fork of f2f called appointments now too. Pull requests very welcome!!!

Issue #85, opened 2022-07-18
Cancellation notice throws error


andrewmadden:
Introduced in https://github.com/catalyst/moodle-mod_facetoface/commit/37011d35205e3325407bcabb62e48b9733cace9f

A superfluous parameter was added to `facetoface_send_notice` in lib.php. `$htmlmessage` parameter is always overwritten. 

All implementations of `facetoface_send_notice` did not implement the new required parameter. 

The parameter `$htmlmessage` should be removed.

```
Exception - Too few arguments to function facetoface_send_notice(), 7 passed in [dirroot]/mod/facetoface/lib.php on line 2196 and exactly 8 expected
```

PR #86, opened 2022-07-18
Fixes issue #85: Remove superfluous parameter from facetoface_send_notice.

andrewmadden:
Fixes bug introduced in commit #37011d35205e.


danmarsden:
thanks Andrew - looks good to me - thanks for fixing the other codechecker things too!

PR #90, opened 2022-09-20
Allow users to join sign up for multiple sessions per activity.

cameron1729:
### Testing
1. Add a face-to-face activity to a course, make sure "Signup type" is set to multiple
2. As a teacher in the course, create two sessions
3. As a student in the course sign up for one of the sessions
4. Check the course home page, ensure the face-to-face activity shows you are signed up for one session, and the other one is available to join
5. Check the "View all sessions" link and confirm it is consistent
6. Sign up for the other session and verify the "View all sessions" page and the course home page show you signed up for both sessions
7. Cancel one of the sign ups to verify you can cancel and rejoin


db/upgrade.php
```diff
@@ -758,5 +758,18 @@ function xmldb_facetoface_upgrade($oldversion=0) {
         upgrade_mod_savepoint(true, 2022031800, 'facetoface');
     }
 
+    if ($oldversion < 2022091900) {
+        $table = new xmldb_table('facetoface');
+        $field = new xmldb_field('signuptype', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, null, '0', 'allowcancellationsdefault');
```
danmarsden, line 763: the length specified here seems different from the one in install.xml.... did you use xmldb editor to do this stuff? :-)

lib.php
```diff
@@ -70,6 +70,10 @@
 define('F2F_CAL_COURSE', 1);
 define('F2F_CAL_SITE',   2);
 
+// Signup setting constants.
```
danmarsden, line 73: This violates the naming conventions on constant names... I know there are other existing constants that don't meet the naming guidelines (and this follows that format) but it would be preferable to not add more.....

db/upgrade.php
cameron1729: Thanks, fixed that one.

lib.php
cameron1729: Changed them to `MOD_FACETOFACE_SIGNUP_...`

PR #93, opened 2022-10-12
Allow users to bulk sign up for every session in the activity

cameron1729:
This patch adds an option to the module create form that changes how students register for sessions. When enabled, instead of one link per session, there is one link per activity, and clicking it signs the student up for every session in the activity.

If a student registers for all sessions using this link, and a teacher later adds a subsequent session, the student will not be automatically registered for that session.


dmitriim:
What about cancelling the sessions (if I signed up for a stream)? should it be cancelled per session or per stream?

lang/en/facetoface.php
```diff
@@ -252,6 +252,10 @@
 $string['moreinfo'] = 'More info';
 $string['multidate'] = '(multi-date)';
 $string['multiple'] = 'Multiple';
+$string['multiplesignupmethod'] = 'Multiple signup method';
+$string['multiplesignupmethod_help'] = 'When set to "single" each session within the activity will have an individual sign up link. When set to "multiple" only one link wil be displayed for the entire activity, any student who registers using this single link will be registered for all sessions within the activity.';
```
dmitriim: After reading this I still not really understand the difference between "Per session" vs "Per activity".

mod_form.php
```diff
@@ -264,4 +273,16 @@ public function data_preprocessing(&$defaultvalues) {
             ];
         }
     }
+
+    public function validation($data, $files) {
```
dmitriim: The following error is a little confusing as setting up to Single hides "Multiple signup method". Should w just aromatically change it or ignore if it's set to singe? 

image

dmitriim: I actually needed to save activity twice to go passed this error

cameron1729: Good point, I've removed this validation and simply forced the activity to be in a valid state when saved.

lang/en/facetoface.php
cameron1729: I've added attempted to make this feature clearer. It's a bit of a challenge since the plugin isn't really supposed to work this way and we are kind of shoe-horning functionality in :s

cameron1729:
Thanks for the review @dmitriim - I've added some fixup commits to address your comments. If these are OK let me know so I can rebase before the merge. Otherwise let me know if there are further changes needed. Cheers.

danmarsden:
no objections, haven't reviewed the functionality but once @dmitriim is happy just merge this in! :-)

mod_form.php
dmitriim: Thank you! works better now

lang/en/facetoface.php
dmitriim: Looks like there is the same help text for "Signup type" and "Multiple signup method" setting. That's why I was complaining about.

image

dmitriim:
When I have Signup type = Multiple and Multiple signup method = Per activity, when as a student I go to an activity itself I can see all sessions listed and sign up option against every session. When I book one session it automatically book me in for all sessions without notifying about that. 

Should we change inside activity UI as well? Possible options:

- Have the same UI as from the course page with only one link "Sign-up for stream"
- After clicking on individual Sign-up link display a notice that you will be signed up for all sessions
- Replace wording in Options column from Sign-up to Sign-up for stream + maybe show a notification?

lang/en/facetoface.php
cameron1729: Ah I see what happened. Actually they are different, but I made a typo by writing "single" and "multiple" instead of "Per session" and "Per activity". I've fixed that now.

cameron1729:
Good catch about that other signup link view @dmitriim - I've pushed a commit implementing option 1

dmitriim:
Thanks @cameron1729 it looks good now. I have created new issue https://github.com/catalyst/moodle-mod_facetoface/issues/94 that can b e solved separately. 
Merging this one in.

PR #95, opened 2022-10-17
Limit candidate attendees to users enrolled in course

alexmorrisnz:


dmitriim:
Hi @alexmorrisnz I don't know a background behind this change, but it looks like it changes an existing logic of building a list of users. Would you consider to add an extra setting to be able to change this logic optionally? That should be relatively easy change that may help people relying on the existing way to be happy.

alexmorrisnz:
Hi @dmitriim I've added a admin level setting to control this.

This has come about as we have a client with 150K+ users who are running into issues with adding the wrong users to sessions as they share names with users not enrolled in the course.

Once this is merged I'll submit two more PRs to add it to 3.11 and 3.4+ branches

lib.php
```diff
@@ -4201,9 +4203,12 @@ public function find_users($search) {
 
         $fields      = 'SELECT ' . $this->required_fields_sql('u');
         $countfields = 'SELECT COUNT(u.id)';
+        $limitsql = "JOIN {user_enrolments} ue ON u.id = ue.userid
+                  JOIN {enrol} e ON e.id = ue.enrolid AND e.courseid = :courseid";
         $sql = "
-                  FROM {user} u
-                 WHERE $wherecondition
+                  FROM {user} u " .
+            (get_config(null, 'facetoface_limit_candidates') ? $limitsql : "") .
```
dmitriim: Do you think the following code snippet looks cleaner?  

```
        $fields      = 'SELECT ' . $this->required_fields_sql('u');
        $countfields = 'SELECT COUNT(u.id)';

        $limitsql = '';
        if (get_config(null, 'facetoface_limit_candidates')) {
            $params['courseid'] = $this->courseid;
            $limitsql = "JOIN {user_enrolments} ue ON u.id = ue.userid
                  JOIN {enrol} e ON e.id = ue.enrolid AND e.courseid = :courseid";
        }

        $sql = "
                  FROM {user} u
                       $limitsql
                 WHERE $wherecondition
                   AND u.id NOT IN
                       (
                       SELECT u2.id
                         FROM {facetoface_signups} s
                         JOIN {facetoface_signups_status} ss ON s.id = ss.signupid
                         JOIN {user} u2 ON u2.id = s.userid
                        WHERE s.sessionid = :sessid
                          AND ss.statuscode >= :statuswaitlisted
                          AND ss.superceded = 0
                       )
               ";
        $order = " ORDER BY u.lastname ASC, u.firstname ASC";
        $params = array_merge($params,
            array(
                'sessid' => $this->sessionid,
                'statuswaitlisted' => MDL_F2F_STATUS_WAITLISTED
            ));

```

alexmorrisnz: Yes that is cleaner, I've pushed it to the branch. Thanks

dmitriim:
@alexmorrisnz can you please create an issue? That would help other people who potentially struggling with the same issue to see that it has been fixed already.

dmitriim:
The change looks good to me, but I'd like @danmarsden to have a look also before we can merge it in.

dmitriim:
@alexmorrisnz probably you also want to bump a version to make sure lang changes are applied.

danmarsden:
+1 from me (after version bump is also added) - feel free to merge that when ready! :-)

dmitriim:
@alexmorrisnz this one is merged. Please back-port to MOODLE_34_STABLE and MOODLE_311_STABLE. Thanks!

PR #96, opened 2022-10-18
Backport signup per activity feature (MOODLE_311_STABLE)

cameron1729:


db/install.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<XMLDB PATH="mod/facetoface/db" VERSION="20220922" COMMENT="XMLDB file for Moodle mod/facetoface."
+<XMLDB PATH="mod/facetoface/db" VERSION="20221018" COMMENT="XMLDB file for Moodle mod/facetoface."
```
dmitriim, line 2: do we care about different version here?

cameron1729: Not sure, but Moodle HQ have got upset in the past when this number doesn't change. I honestly don't think it matters.

dmitriim: I mean should it match the version we are setting in version.php? Probably not...

danmarsden: No it shouldn't they are different things and different lengths.. But yes xmldb editor bumps it when you edit and save so it's almost a validation that the dev used xmldb editor to update the file.

cameron1729: Oh right, no it doesn't need to match the version in version.php - it's a different thing. And actually it's hardly used in code at all.

PR #97, opened 2022-10-18
 Backport signup per activity feature (MOODLE_34_STABLE)

cameron1729:


signup.php
```diff
@@ -204,16 +240,23 @@
     exit;
 }
 
+<<<<<<< HEAD
```
dmitriim: @cameron1729  ooops

cameron1729: :facepalm: fixed

Issue #101, opened 2022-11-08
Add option to add a student to every session in an activity


cameron1729:
When signup type is set to "Multiple" and multiple signup method is set to "Per activity", when a student registers, they are registered for every session in the activity.

When a teacher adds a student via the "Upcoming sessions" screen, they can only add them per session. When signup type is set to "Multiple" and multiple signup method is set to "Per activity" it would make sense to allow the teacher to register the student for every session.

PR #107, opened 2022-12-15
 Backport bulk adding to all sessions (MOODLE_34_STABLE)

cameron1729:
Backport of issue #101


dmitriim:
@cameron1729 looks like we need a version bump to suck in new lang strings

PR #111, opened 2023-01-24
Issue #110: Add control for manager approvals at site level

### LINKED ISSUE #110: Add a plugin setting to allow approvals to be disabled sitewide.
Use case:

Currently, if the manager email settings are set to empty in the plugin settings, a user can still require approvals when they create a new activity. When a student is signing up to a session with approvals enabled, they get a message pointing them to the system administrator. 

Instead, if the manager email is not set up (with the intention to prevent approvals), there should be no option to require approvals when creating a new activity and the student should not see an error message they are unable to action.

Proposed solution:

* New plugin setting that can toggle enabling manager approvals at a site level.
* New setting will affect which settings are displayed on a facetoface mod_form. 
* New setting will be top level control whether a signup should require manager approval.

andrewmadden:
* Add new plugin setting
  * Activities should only allow approvals if enabled at site level


settings.php
```diff
@@ -65,10 +65,17 @@
     0
 ));
 
+$settings->add(new admin_setting_configcheckbox(
+    'facetoface_enableapprovals',
```
danmarsden: I know the other settings are defined in the wrong place, but we shouldn't add more in that global $CFG
can you put this in the correct location facetoface/enableapprovals - defined in config_plugins rather than global $CFG - and one day we might fix the others.

lang/en/facetoface.php
```diff
@@ -235,6 +235,7 @@
 $string['managerchanged'] = 'Your manager\'s email address has been changed.';
 $string['manageremail'] = 'Manager\'s email';
 $string['manageremailaddress'] = 'Manager\'s email address';
+$string['manageremaildisabled'] = 'Manager\'s email settings are not available if manager approvals are not enabled.';
```
keevan, line 238: Could we change this so that the language is in a more positive tone. Unfortunately I tripped a bit on the double negative.

For example, manager email settings are only available when manager approvals are enabled. 

Alternatively, manager email settings are disabled. Manager approvals must be enabled to use this feature.

lang/en/facetoface.php
```diff
@@ -363,6 +364,8 @@
 $string['setting:addchangemanageremail_caption'] = 'Manager\'s email:';
 $string['setting:attendeesexportfields_caption'] = 'Attendees export fields';
 $string['setting:attendeesexportfields'] = 'Select the fields to be included in a session\'s exported list of attendees. This will be in addition to the attendee\'s first and last name.';
+$string['setting:enableapprovals'] = 'When disabled, approvals can not be required for any Face-to-Face activity.';
```
keevan: Cannot -> are not?

lib.php
```diff
@@ -3628,7 +3628,7 @@ function facetoface_print_session($session, $showcapacity, $calendaroutput=false
     // Display requires approval notification.
     $facetoface = $DB->get_record('facetoface', array('id' => $session->facetoface));
 
-    if ($facetoface->approvalreqd) {
+    if (get_config(null, 'facetoface_enableapprovals') && $facetoface->approvalreqd) {
```
keevan: Could we move this to it's own function. The check appears to be called in multiple places and is especially confusing when used in the negative (e.g. not config OR not approval reqd)

Alternatively, the approvalreqd property could be dissected and updated to only be true if the config is set, and then should already be correct in other locations.

How does that sound?

keevan (commented):
A few small notes. I agree with the suggested settings adjustment. I'll have another look once that comes through.

Thanks Andrew

lang/en/facetoface.php
andrewmadden: This message is only to explain why a settings section is empty. The settings api does not allow the setting to be merely disabled which would be a better experience. 

If you can suggest a positive message that still conveys the information, please provide it. I haven't been able to come up with one. 

If it's too confusing, I would just remove this message altogether and have the empty settings section.

lib.php
andrewmadden: The `approvalreqd` setting is per activity instance. It should be not be reset based on the plugin setting `enableapprovals`. If activities are set up to use approvals, and then an unwitting site admin disabled, then re-enabled the `enableapprovals` setting the activity settings should be in the same state as before.

This pattern of check only appears in two places which is why I didn't turn it into a function (Normally use rule of 3), but I can do to make it more readable.

lang/en/facetoface.php
andrewmadden: Incorrect. They do not mean the same thing. When disabled, the option to add approvals via the activity setting is no longer available.

settings.php
andrewmadden: Will update, thanks. Named that on autopilot.

lang/en/facetoface.php
andrewmadden: Some images of settings page if it helps create a better message.

image
Screenshot from 2023-01-27 15-31-03

lib.php
andrewmadden: Added new helper method.

lang/en/facetoface.php
andrewmadden: No action taken.

lang/en/facetoface.php
andrewmadden: No action taken, but happy for a suggestion on rewording this message.

settings.php
andrewmadden: Updated the setting name.

As it was a pretty quick find/replace I also updated the namespaces of the other settings. See issue #116

lib.php
keevan: > This pattern of check only appears in two places which is why I didn't turn it into a function (Normally use rule of 3), but I can do to make it more readable.

Thanks Andrew, I noticed 3 places it could-be/is used, namely, the two places the helper function replaced, and the following function
```
function facetoface_manager_needed($facetoface) {
    return get_config('facetoface', 'enableapprovals')
        && (
            $facetoface->approvalreqd
            || $facetoface->confirmationinstrmngr
            || $facetoface->reminderinstrmngr
            || $facetoface->cancellationinstrmngr
        );
}
```
Previously the logic only applied the inner conditional. And the new check whilst it makes sense, might also take the following form, which should still technically prove true, and be a bit shorter - reasoning being that the approval required goes hand in hand with the helper function.

```
function facetoface_manager_needed($facetoface) {
    return \mod_facetoface\helper::is_approval_required((object) $facetoface)
        && (
            $facetoface->confirmationinstrmngr
            || $facetoface->reminderinstrmngr
            || $facetoface->cancellationinstrmngr
        );
}
if () {
```

andrewmadden: Hi @keevan , I would have simplified, except the logic is not the same for checking whether approvals are required versus if a manager is needed.

There are two scenarios where the above are not equal (I can provide all logic pairs if you want too).

For simplication: a = get_config('facetoface', 'enableapprovals'), b = $facetoface->approvalreqd, c = $facetoface->confirmationinstrmngr || $facetoface->reminderinstrmngr || $facetoface->cancellationinstrmngr

1. a = true, b = false, c = true
  Existing logic: true, New logic: false
2. a = true, b = true, c = false
  Existing logic: true, New logic: false
  
These two methods should remain decoupled as they are checking different states. I can add in unit tests for the existing function `facetoface_manager_needed` if you think it's required?

lang/en/facetoface.php
keevan: Thanks for the screenshots - both help paint the picture clearer. 

The top image looks good. I would prefer having a description under the heading, as opposed to an empty section.

The details under the "Manager Emails" section heading can be omitted. Instead, it'd be ideal if the description was added to the `enableapprovals` control, since that's the trigger point.

lib.php
keevan: Ah, yup you're right.

The odd bit I find from reading this, is the function returns if the manager is needed, but it's dependent on all those factors. Confirmation/Reminder/Cancellation all seem like email notifications, which only works if manager approval is enabled per the new string change.

> $string['manageremaildisabled'] = 'Manager\'s email settings are not available if manager approvals are not enabled.';

Side note: the language between manager approvals being enabled sitewide, vs required for the activity, is subtle and can be confusing.

Scenario 1: 
**A manager is needed** if managers can approve sitewide, approval is not required, and some notification is enabled

It's an odd case but I'm sure it happens. I'm okay with keeping your patch as you've explained the states.

lang/en/facetoface.php
keevan:
```suggestion
$string['setting:enableapprovals'] = 'When disabled, the option to add approvals via the activity setting is no longer available';
```
That's much clearer for me.

lang/en/facetoface.php
andrewmadden: Bit confused about what you are suggesting?

The settings API doesn't allow non editable elements such as a plain text element to be hidden conditionally. So the text explaining why manager's email settings are not there when approvals are disabled will always be shown to user.

> I would prefer having a description under the heading, as opposed to an empty section.

This is the current state. That's how Moodle renders it. The explanation text is the 'description' of that heading.

> The details under the "Manager Emails" section heading can be omitted. Instead, it'd be ideal if the description was added to the enableapprovals control, since that's the trigger point.

I don't want to append it to the description of the checkbox as I believe it would muddy the description of the functionality. 

Do you want me to move it as a text element under the enableapprovals checkbox element? This means the Manager's email section will just be empty with no explanation when the checkbox is ticked.

I thought it made sense to explain what is affecting the section, in the section providing information to the user of what is affecting it. For example if a bunch of settings were added between these elements, and someone checks they box. They may not notice the settings being hidden and then when they scroll down to the section, it will just be empty with no explanation why.

danmarsden:
@keevan you seem to have done most of the PR here so far - do you want to come back and review the last changes to see if it makes sense to merge this in yet?

lang/en/facetoface.php
keevan: > This is the current state. That's how Moodle renders it. The explanation text is the 'description' of that heading.

Yes, this response was targeting a previous response given where you mentioned "If it's too confusing, I would just remove this message altogether and have the empty settings section."

If this is the current state then I'm happy with it.

> I don't want to append it to the description of the checkbox as I believe it would muddy the description of the functionality.

That's a fair point, and as long as that control `enableapprovals` intuitively implies that the section `Manager Emails` is irrelevant, I'm happy for it to not be included. A loose comparison would be a switch, turning off power to a building. You wouldn't need to label "microwave will no longer work" next to the switch, because that's an implied effect from the control. 

Would it make more sense, and be more clear, to relabel that section "Manager approval emails"? (I'm not sure what other settings are under that group but if not then ignore this comment)

lang/en/facetoface.php
andrewmadden: updated

lang/en/facetoface.php
andrewmadden: I'm going to make the call that it is as clear as it can be without wasting more time on this. I'm happy as it is.

PR #122, opened 2023-03-12
Issue #110: Add control for manager approvals at site level (3.11 backport)

andrewmadden:
Backport of https://github.com/catalyst/moodle-mod_facetoface/pull/111


settings.php
```diff
@@ -34,7 +34,7 @@
 require_once($CFG->dirroot . '/user/profile/lib.php');
 
 $settings->add(new admin_setting_configtext(
-    'facetoface_fromaddress',
+    'facetoface/fromaddress',
```
Peterburnett, line 37: Given we are changing all the configs over, I think we just need an upgrade script to grab all configs in the global table for F2F, and write them across to the new table, otherwise settings will be lost on upgrade.

andrewmadden: Added an upgrade script: https://github.com/catalyst/moodle-mod_facetoface/pull/122/files#diff-fcb317575e9e6245fdf1590e51a86685b61c2cc9573a9df8034323ccb3ef614fR787

Issue #116, opened 2023-01-27
Move plugin settings from global namespace to plugin namespace. 


andrewmadden:
E.g. 'facetoface_test' -> 'facetoface/test'

PR #124, opened 2023-03-13
issue #116: Add upgrade script to copy old settings to new settings.

andrewmadden:
Add an extra upgrade script to migrate the settings.


Peterburnett:
This has red CI due to the whitespace decision around = and => from HQ. Merging the feature as its a bug, and we can revisit the CI issue, hopefully at the shared workflow level

PR #129, opened 2023-10-02
Fix confirmationmessage problems

skodak:
* it makes no sense to support files in messages (it triggers debugging during completion reset which breaks behat tests)
* wrong use of trusttext
* better use data_postprocessing
* missing text format code


danmarsden:
Thanks Petr!

PR #135, opened 2023-10-22
Put site wide event creation option behind a capability

### LINKED ISSUE #134: Capability for creating a site event
Hi, would it be possible to add a capability to control the ability to create a f2f activity as a site event? 

we have the odd occasion where activities have been set up as site events by mistake and the result it it displaying in all courses in upcoming events. 

cheers

cameron1729:
Fixes #134

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -101,11 +101,15 @@ class mod_facetoface_mod_form extends moodleform_mod {
 
         $mform->addElement('header', 'calendaroptions', get_string('calendaroptions', 'facetoface'));
 
-        $calendaroptions = array(
+        $calendaroptions = [
             F2F_CAL_NONE   => get_string('none'),
-            F2F_CAL_COURSE => get_string('course'),
-            F2F_CAL_SITE   => get_string('site')
-        );
+            F2F_CAL_COURSE => get_string('course')
+        ];
+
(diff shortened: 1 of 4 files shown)


mod_form.php
```diff
@@ -101,11 +101,11 @@ public function definition() {
...
         $mform->addElement('header', 'calendaroptions', get_string('calendaroptions', 'facetoface'));
 
-        $calendaroptions = array(
+        $calendaroptions = [
             F2F_CAL_NONE   => get_string('none'),
-            F2F_CAL_COURSE => get_string('course'),
-            F2F_CAL_SITE   => get_string('site')
-        );
+            F2F_CAL_COURSE => get_string('course')
+        ] + (has_capability('mod/facetoface:createsitewideevent', $this->context) ? [F2F_CAL_SITE   => get_string('site')] : []);
```
keevan: This is a bit hard to read, maybe split out the ternary to a variable?

keevan: Actually, it would be clearer to read as a basic if condition after the array.

e.g.
```
array = [ 2 options ] 
if (condition) {
array[key] = value
}
```

cameron1729: Thanks @keevan - updated the patch.

Issue #137, opened 2024-02-15
Plugin setting session_roles in plugin namespace are not being used in lib.php


joshwillcock:
In issue  #116 the plugin settings were moved from the global namespace to the plugin namespace.

When checking for available settings roles it is checking the old $CFG location meaning any saved results are not loaded.

Simple/quick fix attached.

Thanks.

Josh

PR #138, opened 2024-02-15
#137 Plugin setting session_roles in plugin namespace not being used.

joshwillcock:
In issue https://github.com/catalyst/moodle-mod_facetoface/issues/116 the plugin settings were moved from the global namespace to the plugin namespace.

When checking for available settings roles it is checking the old $CFG location meaning any saved results are not loaded.

Simple/quick fix attached.

Thanks.

Josh


danmarsden:
Thanks @joshwillcock !

PR #143, opened 2024-03-02
Fixed Behat tests

### LINKED ISSUE #142: 4.3 behat errors
Just picking out the errors that were noted in #141 
```
        Exception: debugging() message/s found:
        Invalid event functions defined in \mod_facetoface\event\course_module_viewed
        line 275 of /lib/classes/event/base.php: call to debugging()
        line 86 of /mod/facetoface/view.php: call to core\event\base::create()
```

https://github.com/catalyst/moodle-mod_facetoface/blob/MOODLE_403_STABLE/view.php#L86
https://github.com/catalyst/moodle-mod_facetoface/blob/MOODLE_403_STABLE/classes/event/course_module_viewed.php#L41
https://moodledev.io/docs/4.2/devupdate#validation-of-required-parameter

probably needs to be more explicitly declared
https://github.com/mudrd8mz/moodle-tool_pluginskel/blob/main/skel/file/mod/view.mustache#L59-L62
not sure. Not an error we've seen in any other upgrade prep so far

fyi @opitz

opitz:
* removed deprecated legacy methods
* updated Behat tests due to interface changes in Moodle 4.x

This will fix #142.

### DIFF
--- a/classes/event/add_session.php
+++ b/classes/event/add_session.php
@@ -80,16 +80,6 @@ class add_session extends \core\event\base {
     }
 
     /**
-     * Return the legacy event log data.
-     *
-     * @return array|null
-     */
-    protected function get_legacy_logdata() {
-        return array($this->courseid, $this->objecttable, 'add session', 'sessions.php?s=' . $this->objectid,
-            $this->objectid, $this->contextinstanceid);
-    }
-
-    /**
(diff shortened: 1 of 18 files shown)


danmarsden:
awesome - thanks heaps for the PR! :-)

PR #151, opened 2024-05-01
feat: bulk book users to sessions via csv file

### LINKED ISSUE #150: Be able to upload bookings (link users to a session in bulk) via csv upload

keevan:
Summary:
- permission check to only allow users with this permission to do so
- can intially set notification type, status (booked/cancelled) and
  discount code.
- tests for various scenarios.
- managed via a booking manager class.
- supports both csv file and array (list of bookings) formats.

Resolves #150

TODO: 
- [x] add event handling(?)

### DIFF
--- a/classes/booking_manager.php
+++ b/classes/booking_manager.php
@@ -1,0 +1,314 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace mod_facetoface;
+
+use context_course;
+use moodle_exception;
+use lang_string;
+
+class booking_manager {
+
+    /** @var stored_file the file to process as a stored_file object */
+    private $file;
+
+    /** @var int The facetoface module ID. */
+    private $f;
+
+    /** @var int The course id. */
+    private $course;
+
+    /** @var int The course id. */
+    private $facetoface;
+
+    /** @var array collection of records (if loaded from memory), in an array. */
+    private $records;
+
+    /** @var bool Whether or not the bookings are loaded from a file. */
+    private $usefile = true;
+
+    /**
+     * Constructor for the booking manager.
+     * @param int $f The facetoface module ID.
+     * @param array $records The records to process.
+     */
+    public function __construct($f, $records = []) {
+        global $DB;
+
+        if (!$facetoface = $DB->get_record('facetoface', ['id' => $f])) {
+            throw new moodle_exception('error:incorrectfacetofaceid', 'facetoface');
+        }
+        if (!$course = $DB->get_record('course', ['id' => $facetoface->course])) {
+            throw new moodle_exception('error:coursemisconfigured', 'facetoface');
+        }
+
+        $this->f = $f;
+        $this->facetoface = $facetoface;
+        $this->course = $course;
+        $this->records = $records;
+    }
+
+    /**
+     * Returns file from file system. File must exist.
+     * @param int $fileitemid Item id of file stored in the current $USER's draft file area
+     * @return stored_file
+     */
+    public function load_from_file(int $fileitemid) {
+        global $USER;
+        $this->usefile = true;
+
+        $fs = new file_storage();
+        $files = $fs->get_area_files(context_user::instance($USER->id)->id, 'user', 'draft', $fileitemid, 'itemid', false);
+
+        if (count($files) != 1) {
+            throw new moodle_exception('error:cannotloadfile', 'mod_facetoface');
+        }
+
+        return current($files);
+    }
+
+    /**
+     * Load in the records to process from an array
+     * @param array $records
+     */
+    public function load_from_array(array $records) {
+        $this->usefile = false;
+        $this->records = $records;
+
+        return $this;
+    }
+
+    public static function get_headers(): array {
+        return [
+            'email',
+            'session',
+            'status',
+            'discountcode',
+            'notificationtype',
+        ];
+    }
+
+    private function get_iterator() {
+        if (!$this->usefile) {
+            foreach ($this->records as $record) {
+                yield $record;
+            }
+            return;
+        }
+
+        $handle = $this->file->get_content_file_handle();
+        $maxlinelength = 1000;
+        $delimiter = ',';
+        $rownumber = 1; // First row is headers.
+        $headers = self::get_headers();
+        $numheaders = count($headers);
+        fgets($handle); // Move pointer past first line (headers).
+        try {
+            while (($data = fgetcsv($handle, $maxlinelength, $delimiter)) !== false) {
+                $rownumber++;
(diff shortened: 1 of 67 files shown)


keevan:
Pulled in some changes found in 403, goal is that the behat issues have already been resolved.

classes/booking_manager.php
```diff
@@ -0,0 +1,316 @@
...
+
+            // Set defaults for fields with no value.
+            $entry->status = $entry->status ?? '';
+            $entry->notificationtype = $entry->notificationtype ?? '';
+            $entry->discountcode = $entry->discountcode ?? '';
+
+            // Validate and get user.
+            $userids = $DB->get_records('user', ['email' => $entry->email], 'id');
+
+            // Multiple matched, ambiguous which is the real one.
```
Peterburnett, line 180: Given that these have a couple of conditions on most of the fields, I reckon we should just split this up into a couple of private functions that just take a field -> errors, or row -> errors if there is cross dependencies. that way the main loop just reads as the series of checks being done at a high level

classes/form/confirm_bookings_form.php
```diff
@@ -0,0 +1,67 @@
...
+        $mform->addElement('hidden', 'f');
+        $mform->setType('f', PARAM_INT);
+
+        // Reference to the uploaded file.
+        $mform->addElement('hidden', 'fileid', $fileid);
+        $mform->setType('fileid', PARAM_INT);
+
+        $backurl = new moodle_url('/mod/facetoface/upload.php', ['f' => $f]);
+        $htmlbuttons = $OUTPUT->render((new single_button(
+            new moodle_url('upload.php', ['f' => $f, 'fileid' => $fileid, 'process' => 1]),
```
Peterburnett: one is an absolute URL, one is a relative, make these consistent?

classes/form/upload_bookings_form.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @package    mod_facetoface
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright  Catalyst IT, 2024
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class upload_bookings_form extends \moodleform {
+
+    /**
+     * Build form for importing woekflows.
```
Peterburnett: typo in comment

classes/booking_manager.php
keevan: I tested for user, and that was okay.., being that it worked when I added a method like `check_user($email, &$errors, $row) -> $userid`, but the problem is it introduced some level of inconsistency in how the checks are performed and recorded compared to sessions, which required different parameters to be provided, along with a session cache object, so it's probably better to keep all the checks inline for now, until there's a better path forward for all the checks collectively.

classes/form/upload_bookings_form.php
```diff
@@ -0,0 +1,73 @@
...
+     * @see \moodleform::definition()
+     */
+    public function definition() {
+        global $CFG;
+
+        $mform = $this->_form;
+
+        $mform->addElement('header', 'settingsheader', get_string('upload'));
+
+        $url = new moodle_url('example.csv');
```
Peterburnett: Not sure if there is a convention or code style guide on this, but generally im partial to absolute URLs where we can, so the PHP file/function can be moved around without invisibly breaking a link to a static asset

keevan: I blatantly copied the style from here:
https://github.com/moodle/moodle/blob/master/admin/tool/uploaduser/user_form.php#L43

But I can change it to absolute if preferred.

upload.php
```diff
@@ -0,0 +1,135 @@
...
+    throw new moodle_exception('error:incorrectfacetofaceid', 'facetoface');
+}
+if (!$course = $DB->get_record('course', ['id' => $facetoface->course])) {
+    throw new moodle_exception('error:coursemisconfigured', 'facetoface');
+}
+if (!$cm = get_coursemodule_from_instance('facetoface', $facetoface->id, $course->id)) {
+    throw new moodle_exception('error:incorrectcoursemoduleid', 'facetoface');
+}
+
+
```
Peterburnett: double newline here

upload.php
```diff
@@ -0,0 +1,135 @@
...
+require_course_login($course, true, $cm);
+$context = context_course::instance($course->id);
+$modulecontext = context_module::instance($cm->id);
+require_capability('mod/facetoface:editsessions', $context);
+require_capability('mod/facetoface:uploadbookings', $context);
+
```
Peterburnett, line 56: another one

upload.php
```diff
@@ -0,0 +1,135 @@
...
+    $bm->load_from_file($fileid);
+
+    // Validate entries.
+    $errors = $bm->validate();
+    if (empty($errors)) {
+        // Process entries.
+        $bm->process();
+
+        echo "PROCESSED";
+        die;
```
Peterburnett: leftover die?

upload.php
```diff
@@ -0,0 +1,135 @@
...
+        // Redirect back to start with notification.
+        redirect(
+            new moodle_url('/mod/facetoface/upload.php', ['f' => $f]),
+            'Successfully processed X records.',
```
Peterburnett: This bit looks like it wants to be a parameterised lang string?

upload.php
keevan: Good catch

upload.php
```diff
@@ -0,0 +1,135 @@
...
+
+    $errmsg = get_string('error:bookingsuploadfileerrorsfound', 'mod_facetoface', count($errors));
+    redirect(
+        new moodle_url('/mod/facetoface/upload.php', ['f' => $f]),
+        $errmsg,
+        null,
+        notification::NOTIFY_ERROR);
+} else {
+    $mform = new upload_bookings_form(null);
+    $data = $mform->get_data();
```
Peterburnett: is this get_data call actually needed?

upload.php
keevan: Whoops.

upload.php
keevan: I made it a bit more general. The number of records processed would equal the number of rows in the file. I don't think that information is important as it's either all or nothing, as it wouldn't redirect if there was an exception during processing.

classes/booking_manager.php
keevan: For now I've left it inline and added a comment to the function (which was missing at the time)

danmarsden:
I'm a bit confused here - is this a new feature? - or just a backport? 

new features should land in the latest branch first - then we can deal with the backports?

keevan:
> I'm a bit confused here - is this a new feature? - or just a backport?
> 
> new features should land in the latest branch first - then we can deal with the backports?

Good catch, I'll re-target the latest branch and create a backport PR once it's merged. Thanks Dan.

Peterburnett (approved):
I've largely reviewed the major commit, rather than the CI changes, but this one is looking good to go.

Peterburnett:
@danmarsden I've approved this one, its fairly self contained, did you have anything you wanted to mention before merge?

danmarsden:
if you've reviewed the full patch...and hard-coded english lang strings are fixed like this one I spotted:
https://github.com/catalyst/moodle-mod_facetoface/pull/151/files#diff-2359f56e8ab786049c0f9961df9bd8363b0f1e4e986972ebc692302ebc920000R119

then feel free to merge. :-)

db/access.php
```diff
@@ -30,160 +30,170 @@
...
             'editingteacher' => CAP_ALLOW,
-            'manager' => CAP_ALLOW
-        ]
-    ]
-);
+            'manager' => CAP_ALLOW,
+        ],
+    ],
+
+    // Ability to add a new facetoface activity to a course.
```
Peterburnett: Comment needs updating

keevan: Good spot. Fixed

Peterburnett (approved):

PR #152, opened 2024-05-06
add behat session generator

skodak:


skodak:
hi @danmarsden ,

I have ported some behat generator code from our temporary fork to simplify our migration. I hope it would be useful here too.

danmarsden:
merged - thanks @skodak !

PR #163, opened 2024-05-23
fix: update list to only show session id if user has capabilities to upload

### LINKED ISSUE #161: Session id/number should only be visible to those who use it
e.g. for users with permission to upload and update bulk session attendance via csv

It currently displays it to students as well, which is unintended.

keevan:
Resolves #161

### DIFF
--- a/renderer.php
+++ b/renderer.php
@@ -33,7 +33,14 @@ class mod_facetoface_renderer extends plugin_renderer_base {
     /**
      * Builds session list table given an array of sessions
      */
-    public function print_session_list_table($customfields, $sessions, $viewattendees, $editsessions, $signuplinks = true) {
+    public function print_session_list_table(
+        $customfields,
+        $sessions,
+        $viewattendees,
+        $editsessions,
+        $signuplinks = true,
+        $uploadbookings = false
+    ) {
         $output = '';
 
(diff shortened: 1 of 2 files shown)


renderer.php
```diff
@@ -33,7 +33,14 @@ class mod_facetoface_renderer extends plugin_renderer_base {
     /**
      * Builds session list table given an array of sessions
      */
-    public function print_session_list_table($customfields, $sessions, $viewattendees, $editsessions, $signuplinks = true) {
```
matthewhilton, line 36: I know it probably wasn't there before, but would be nice to add phpdoc for these params, so for e.g. we know what `$uploadbookings` is for

Also if `$uploadbookings` is coming from a capability check, could it maybe be called `$canuploadbookings` to be clearer?

keevan: Yes to the docs, (note that phpdoc checks are disabled because there's a lot of issues in this plugin for that), and I think if we do the $candothing instead of $dothing approach (which I prefer btw), we should apply it consistently throughout, which I'd like to be handled in a separate issue.

matthewhilton: Ah I see the var is already called $uploadbookings before in the code. I guess its alright then.

PR #167, opened 2024-05-27
Fix #54 - remove event name limit

### LINKED ISSUE #54: increasing (or removing) CALENDAR_MAX_NAME_LENGTH ?
Hi,

Is there a particular reason to limit event name to 15 characters ? It is really short.
https://github.com/catalyst/moodle-mod_facetoface/blob/9a177d52908a710ce47ba79b05b36b402f827bf6/lib.php#L68
https://github.com/catalyst/moodle-mod_facetoface/blob/9a177d52908a710ce47ba79b05b36b402f827bf6/lib.php#L3206

AFAIK technically mdl_event.name is a `longtext` and module name which is used to build the event name is a `varchar(255)`.  
I know that you can use "shortname" if you want to display a specific (and length unlimited) event name but it forces teachers to do double entry if they only want the module name to be displayed completely.  

IMO truncate should be done on display only (and is correctly handle by the calendar) and not during event creation.

ak4t0sh:
As stated in #54 

> AFAIK technically mdl_event.name is a longtext and module name which is used to build the event name is a varchar(255).
> I know that you can use "shortname" if you want to display a specific (and length unlimited) event name but it forces teachers to do double entry if they only want the module name to be displayed completely.
> 
> IMO truncate should be done on display only (and is correctly handle by the calendar) and not during event creation.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -65,7 +65,6 @@ define('CUSTOMFIELD_TYPE_SELECT',      1);
 define('CUSTOMFIELD_TYPE_MULTISELECT', 2);
 
 // Calendar-related constants.
-define('CALENDAR_MAX_NAME_LENGTH', 15);
 define('F2F_CAL_NONE',   0);
 define('F2F_CAL_COURSE', 1);
 define('F2F_CAL_SITE',   2);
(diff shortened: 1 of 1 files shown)


danmarsden:
thanks for the PR!

PR #168, opened 2024-07-03
Bulk upload suppress emails

matthewhilton:
**Changes**

- Adds a checkbox to suppress emails when bulk uploading, to bring it in alignment with the manual assignment page.
- A few fixups to the code that were causing debugging messages:
    - Output before redirect
    - Context being set when it didn't need to be (and was trying to set a different context)
    - Undefined var `$courseid`
    - Undefined var `$session`
    - Removed unused import

**Testing**

- Manual testing - works as expected.
- Also covered by unit tests

### DIFF
--- a/classes/booking_manager.php
+++ b/classes/booking_manager.php
@@ -53,6 +53,9 @@ class booking_manager {
     /** @var bool Whether or not the bookings are loaded from a file. */
     private $usefile = true;
 
+    /** @var bool When true, confirmation emails are not sent. */
+    private $suppressemail = false;
+
     /**
      * Constructor for the booking manager.
      * @param int $f The facetoface module ID.
@@ -357,7 +360,7 @@ class booking_manager {
                         $this->transform_notification_type($entry->notificationtype),
                         $statuscode,
                         $user->id,
-                        true
+                        !$this->suppressemail,
                     );
 
                     continue;
@@ -389,5 +392,12 @@ class booking_manager {
         }
 
         return true;
+    }
+
+    /**
+     * Stops confirmation emails from being sent
+     */
+    public function suppress_email() {
+        $this->suppressemail = true;
     }
 }
--- a/upload.php
+++ b/upload.php
@@ -30,7 +30,6 @@ use core\output\notification;
 use mod_facetoface\form\upload_bookings_form;
 use mod_facetoface\form\confirm_bookings_form;
 use mod_facetoface\booking_manager;
-use mod_facetoface\event\csv_processed;
 
 $f = optional_param('f', 0, PARAM_INT); // The facetoface module ID.
 $fileid = optional_param('fileid', 0, PARAM_INT); // The fileid of the file uploaded.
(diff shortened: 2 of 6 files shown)


classes/booking_manager.php
```diff
@@ -53,6 +53,9 @@ class booking_manager {
     /** @var bool Whether or not the bookings are loaded from a file. */
     private $usefile = true;
 
+    /** @var bool Wether the confirmation emails should not be send */
```
keevan:
```suggestion
    /** @var bool Whether the confirmation emails should not be sent */
```
OR (to avoid mental gymnastics with double negatives)
```suggestion
    /** @var bool When true, confirmation emails will not be sent */
```

classes/booking_manager.php
```diff
@@ -390,4 +393,12 @@ public function process() {
 
         return true;
     }
+
+    /**
+     * Sets if the confirmation emails should be suppressed
+     * @param bool $val
+     */
+    public function set_suppress_email($val) {
+        $this->suppressemail = $val;
```
keevan: It might not hurt to add some additional hints, and clarify what the value would be without the phpdocs.
```suggestion
    public function set_suppress_email(bool $suppressed) {
        $this->suppressemail = $suppressed;
```

upload.php
```diff
@@ -97,6 +95,11 @@
     $bm = new booking_manager($f);
     $bm->load_from_file($fileid);
 
+    // Get the options selected by the user at confirm time.
+    $confirmdata = (new confirm_bookings_form(null))->get_data();
+    $suppressemail = !empty($confirmdata->suppressemail);
+    $bm->set_suppress_email($suppressemail);
```
keevan: I wonder if it's generally a good idea to have these setters `set_suppress_email($bool)`, or to have an explicit `$bm->suppress_emails()` method instead, and have the caller control the intended behaviour, but the class control the value being set. 

Wondering in terms of debugging, since it defaults (in the bm class) to send emails, the only times it shouldn't is when it hits this `suppress_emails` call, which can be more clearly observed if viewed via a stack trace for example, instead of needing to see it hit this setter, and checking if the value passed to this setter was indeed true/false and correctly set (or something to that nature).

keevan (commented):
Overall the changes here looks good. Thanks for adding in the tests.

Are you able to consider the feedback comments and once ready, add a commit to also bump the version. 

Also noting that ideally this would land in 403 stable before it is backported to 400. Perhaps we should merge it in that order, if you could prepare a PR for that.

Thanks @matthewhilton.

upload.php
matthewhilton: I think both ways are more or less equal, with my deciding factor being simplicity of `->suppress_emails()` without any var, just setting it to true. Have updated the PR accordingly.

matthewhilton:
Confirming I have rebased and changed the target to `MOODLE_403_STABLE`

Note - After locally testing this, I found a new error: `Exception - Undefined constant "mod_facetoface\form\FILE_INTERNAL"` when loading the form to upload the csv.
This seems to be caused by the reordering of the `$OUTPUT->header` (i.e. it was being implicitly included before). I have added another commit to explicitly include these files as well.

keevan (approved):
Looks good. Thanks for making the changes Matt.

PR #174, opened 2024-07-09
feat: suppress cancel emails in bulk upload

matthewhilton:
**Changes**
- An extension on the suppress bulk upload emails feature added in https://github.com/catalyst/moodle-mod_facetoface/pull/168
- Now also suppresses cancel emails, since they are handled separately from signup emails.

**Testing**
- Manual testing - passed
- Covered by updated unit tests


keevan (approved):
Looks good, thanks @matthewhilton

PR #175, opened 2024-07-09
feat: suppress cancel emails in bulk upload

matthewhilton:
Cherry pick of https://github.com/catalyst/moodle-mod_facetoface/pull/174

### DIFF
--- a/version.php
+++ b/version.php
@@ -30,8 +30,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2023101905;
-$plugin->release   = 2022101905;
+$plugin->version   = 2023101906;
+$plugin->release   = 2023101906;
 $plugin->requires  = 2022031500;  // Requires 4.0.
 $plugin->component = 'mod_facetoface';
 $plugin->maturity  = MATURITY_STABLE;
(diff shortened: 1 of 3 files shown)


version.php
```diff
@@ -30,8 +30,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2023101905;
-$plugin->release   = 2022101905;
+$plugin->version   = 2023101906;
```
keevan, line 33: Looks like the year was bumped. This I believe should be a microbump. 

Once this is fixed it should be good to go.

matthewhilton: ah good catch! updating....

matthewhilton: Actually I think this is correct, the version is 1 yr ahead of the release, these diverged ~ 1yr ago in https://github.com/catalyst/moodle-mod_facetoface/commit/c27e172247dfa3f5fe1128d1f78934db7f6a3bbd

Hence version and release have only had microbump here

keevan: Oh that's very strange. Since release doesn't actually do anything, can you make it equivalent for now to keep it generally consistent for our plugins, and to prevent it raising flags in a future review? I think it's fine to fix it now.

matthewhilton: Yep agreed - updated

Issue #176, opened 2024-08-13
Standardise email formatting


matthewhilton:
Currently the formatting of emails is all over the place:

- Fields such as `confirmationmessage` are editor form elements (hence are html), but the rest of the email related fields are just plaintext with no format.
- Some functions use `facetoface_send_notice` but some don't leading to inconsistencies.
- There is multiple layers/combinations of `format_text`, `nl2br`, `html_to_text` being applied that tends to mangle the output of any replacements

danmarsden:
Thanks @matthewhilton this has annoyed me for a while too so really nice to see it getting a tidy up!

PR #177, opened 2024-08-13
[#139] Ensure all emails send same message

### LINKED ISSUE #139: Cancellation emails incorrectly send plaintext as html, leading to email clients incorrectly formatting them
Imagine you have the cancellation email text set as the following:

```
Hello
Your booking has been cancelled
Thank you
```

If you were to cancel this booking, the user would get two emails (which is a different issue, see https://github.com/catalyst/moodle-mod_facetoface/issues/60)

The email without the .ics attachment will show normally, i.e. what is above. However, the email with the .ics attachment would be joined into a single line, e.g.:
```
Hello Your booking has been cancelled Thank you
```
This looks to be because when sending cancellation emails, when sending the .ics email here:

https://github.com/catalyst/moodle-mod_facetoface/blob/10d6966c09620d76749bf3ca0f4a6245923808bb/lib.php#L2260-L2261

Both the `$messagehtml` and `$messagetext` are set as plaintext (no html), i.e.:
```
Hello
Your booking has been cancelled
Thank you
```
Which leads to email clients interpreting this message as html, but since there are no line breaks, etc... it joins it all into a single one line string.

This does not seem to happen with confirmation messages, because they use a html editor and probably format the text into both html and plaintext properly, instead of the cancellation messages which just use the normal plain textarea editor.

matthewhilton:
Closes #139 

**Changes:**
- Ensures that regardless if you send ical, text, or both the contents of the email received is the same. Previously, the ical email would send the plaintext as html leading to clients showing everything on a single line.
- This is done by only passing in text to the `email_to_user` function, and allowing it to convert it to the appropriate html. This fixes a lot of the missing newline issues.

**Note:** This PR does not attempt to cleanup / fix formatting issues in email. It purely attempts to ensure that the emails sent (ical or not) are the same and fix the newline issue. Fixing email formatting is covered by https://github.com/catalyst/moodle-mod_facetoface/issues/176

**Testing**
- Covered by new unit tests (previously were no email related unit tests)

Manual testing:

| Email | Before | After |
|--------|--------|--------|
| Confirmation iCal | image | image |
| Cancellation | image | image |

**To Do:**
- [x] Ci upstream fix - waiting for upstream nvm issue to be resolved
- [x] Cherry pick to latest branch https://github.com/catalyst/moodle-mod_facetoface/pull/178
- [x] Test manually that emails are formatted decently

### DIFF
--- a/classes/booking_manager.php
+++ b/classes/booking_manager.php
@@ -168,7 +168,10 @@ class booking_manager {
         global $DB;
         $errors = [];
         $sessioncapacitycache = [];
-        $timenow ??= time();
+
+        if ($timenow == null) {
+            $timenow = time();
+        }
 
         // Break into rows and validate the multiple interdependant fields together.
         foreach ($this->get_iterator() as $index => $entry) {
--- a/lib.php
+++ b/lib.php
@@ -2238,6 +2238,12 @@ function facetoface_send_notice($postsubject, $posttext, $posttextmgrheading,
         }
     }
 
+    // Define a simple reusable function so we don't have to copy
+    // and paste a huge function call multiple times.
+    $substitute = function($text, $session) use ($facetoface, $user) {
+        return facetoface_email_substitutions($text, format_string($facetoface->name), $facetoface->reminderperiod, $user, $session, $session->id);
+    };
+
     // Do iCal attachement stuff.
     $icalattachments = [];
     if ($notificationtype & MDL_F2F_ICAL) {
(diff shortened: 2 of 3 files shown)


keevan (commented):
```
The function str_contains() is not present in PHP version 7.4 or earlier
(PHPCompatibility.FunctionUse.NewFunctions.str_containsFound)
```
I checked and 4.1 has a min version of php 7.4, whereas Moodle 4.0 has a min version of php 7.3, so str_contains might not be a compatible function for this branch. 

Similarly for `??=`
```
The null coalesce equal operator (??=) is not present in PHP version 7.3 or earlier
(PHPCompatibility.Operators.NewOperators.t_coalesce_equalFound)
```
and the `fn` keyword
```
The "fn" keyword for arrow functions is not present in PHP version 7.3 or earlier
(PHPCompatibility.Keywords.NewKeywords.t_fnFound)
```

Kind of annoying, and this wouldn't have been an issue for a MOODLE_401_STABLE branch. 

Otherwise, tests look good, changes look good.

danmarsden:
Just create a 4.1 branch, we don't officially support 4.0 now anyway.

matthewhilton:
Thanks @keevan good pickups, i've added the php version compatibility checker to my tooling now as well.

Deciding between forking a new branch and fixing up the issues, I decided to just fix up the issues manually since they were tiny and easy to tweak. Making a whole new branch seems unnecessary to me in this case.

keevan (approved):

PR #184, opened 2024-10-10
fix #183 add space before line break in ICAL document

### LINKED ISSUE #183: Missing space between words on .ics attachment files
It looks like there is a conflict between the use of the method wordwrap() with the cut_long_words parameter set to true and the calendar data format. 
Following this => https://datatracker.ietf.org/doc/html/rfc5545#section-3.1 _"Unfolding
   is accomplished by removing the CRLF and the linear white-space
   character that immediately follows."_

Adding a space before the line break can probably fix it. This is what the method wordwrap would do if the parameter cut_long_words is set to false.

https://github.com/catalyst/moodle-mod_facetoface/blob/MOODLE_403_STABLE/lib.php#L3151

gbarat87:
Closes #183

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -3148,7 +3148,7 @@ function facetoface_ical_escape($text, $converthtml=false) {
     );
 
     // Text should be wordwrapped at 75 octets, and there should be one whitespace after the newline that does the wrapping.
-    $text = wordwrap($text, 75, "\n ", true);
+    $text = wordwrap($text, 75, " \n ", true);
 
     return $text;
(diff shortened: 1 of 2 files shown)


matthewhilton:
Hey @gbarat87 thanks for the PR
I edited the PR description so that it auto-closes and links the issue #183 
In the future, more details in the PR description e.g. what testing was done, reasons why this was done, etc.. would be ideal, however in this case it is such a tiny change its not a blocker.

PR #187, opened 2024-12-02
Validate user sessions on bulk upload when f2f type is single

rhell4:
Currently when using the bulk upload feature a user can be added to multiple sessions when the f2f signup type is set to single.

This adds validation checking on both existing signed up sessions and multiple sessions for a single user in a bulk upload.

Testing:

- F2F activity with signup type set to single
- 3 sessions added to the F2F
- Single student enrolled in the course
- Signup the student to session 1
- Upload a CSV to bulk signup student to session 2 and 3
- You should see the following errors
    - Student is already signed-up for this Face-to-Face activity.
    - User student has more than one session
- Edit F2F and change the signup type to multiple
- Try uploading the same CSV file again
- You should see "Uploaded file has been validated and ready to be processed."
- Click "Confirm and process"
- You should see "Successfully processed records."
- The student should be signed up for all 3 sessions now

### DIFF
--- a/classes/booking_manager.php
+++ b/classes/booking_manager.php
@@ -172,6 +172,8 @@ class booking_manager {
         $sessioncapacitycache = [];
         $timenow ??= time();
 
+        $usersessions = [];
+
         // Break into rows and validate the multiple interdependant fields together.
         foreach ($this->get_iterator() as $index => $entry) {
             $row = $index + 1;
@@ -241,6 +243,21 @@ class booking_manager {
                     $sessioncapacitycache[$session->id]['capacity']--;
                     $sessioncapacitycache[$session->id]['rows'][] = $row;
                 }
+
+                // Don't allow users to signup to another session if the signup type is not multiple.
+                if (isset($userid) && $entry->status !== 'cancelled' && $this->facetoface->signuptype != MOD_FACETOFACE_SIGNUP_MULTIPLE) {
+                    if ($currusersessions = facetoface_get_user_submissions($this->f, $userid)) {
+                        foreach ($currusersessions as $currusersession) {
+                            if ($currusersession->sessionid != $session->id) {
+                                $errors[] = [
+                                    $row,
+                                    new lang_string('error:addalreadysignedupattendee', 'mod_facetoface', $entry->email),
+                                ];
+                                break;
+                            }
+                        }
+                    }
+                }
             }
 
             // Check user enrolment into the course.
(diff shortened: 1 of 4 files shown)


classes/booking_manager.php
```diff
@@ -241,6 +244,20 @@ public function validate($timenow = null): array {
                     $sessioncapacitycache[$session->id]['capacity']--;
                     $sessioncapacitycache[$session->id]['rows'][] = $row;
                 }
+
+                if (isset($userid) && $session && $this->facetoface->signuptype != MOD_FACETOFACE_SIGNUP_MULTIPLE) {
```
matthewhilton: the `&& $session` here is redundant I think, because you are inside a block `if ($session)`

classes/booking_manager.php
```diff
@@ -273,6 +290,35 @@ public function validate($timenow = null): array {
...
+                if (!isset($usersessions[$userid])) {
+                    $usersessions[$userid] = [];
+                }
+
+                $userrows[$userid][$row] = $entry->email;
+
+                if ($session && !in_array($session->id, $usersessions[$userid])) {
+                    $usersessions[$userid][] = $session->id;
+                }
+            }
```
matthewhilton: Is there a simpler way to do this? Maybe we can just store the count instead of the sessions + user rows. Just seems a bit excessive to use two multi level arrays.

If it were me, I think I would break this entire logic out into a new loop to keep it more readable, to avoid growing the current loop with more and more conditions.

classes/booking_manager.php
```diff
@@ -241,6 +244,20 @@ public function validate($timenow = null): array {
                     $sessioncapacitycache[$session->id]['capacity']--;
                     $sessioncapacitycache[$session->id]['rows'][] = $row;
                 }
+
```
matthewhilton, line 246: a comment here about what the block below does would be good

classes/booking_manager.php
```diff
@@ -273,6 +290,35 @@ public function validate($timenow = null): array {
...
+                $userrows[$userid][$row] = $entry->email;
+
+                if ($session && !in_array($session->id, $usersessions[$userid])) {
+                    $usersessions[$userid][] = $session->id;
+                }
+            }
+        }
+
+        if ($this->facetoface->signuptype != MOD_FACETOFACE_SIGNUP_MULTIPLE) {
```
matthewhilton, line 314: a comment about what this block is doing would be good

classes/booking_manager.php
```diff
@@ -273,6 +290,35 @@ public function validate($timenow = null): array {
...
+            foreach ($usersessions as $userid => $sessions) {
```
matthewhilton: Potential refactor idea, that IMO would be more readable:

```
$doublebookedsessions = array_filter($usersessions, fn($sessions) => return count($sessions) > 1, ARRAY_FILTER_USE_KEY)
$doublebookederrors = array_map(fn($userid) => [...build error here], $doublebookedsessions)
$errors = array_merge($errors, $doublebookederrors);
```

classes/booking_manager.php
rhell4: You're right, I think I added it when I did the if statement further down. I've removed that check but I've also put in a `$entry->status !== 'cancelled'` which I overlooked.

classes/booking_manager.php
rhell4: The reason I didn't do a simple count is because multiple rows for the same session is technically valid, so we need to know which distinct sessions the user is being added to. We also need the details of the email and rows so that we can create the errors. I have reworked this though so it's more like the session over capacity validation.

classes/booking_manager.php
rhell4: Added a comment.

classes/booking_manager.php
rhell4: I've added some comments to explain what is happening here.

classes/booking_manager.php
rhell4: I've changed this a bit along with the `$usersessions` array structure so it matches how the over capacity sessions validation works. Let me know what you think.

classes/booking_manager.php
matthewhilton: Yep that makes sense :+1:

Issue #205, opened 2025-05-05
Activity front screen changes


golenkovm:
Business requirements: Location should be a visible field on the front screen.

Implementation approach:
- Add config in settings.php (dropdown/select element for a custom field). This would be a dropdwon/select element populated with customfields from the facetoface_list_of_customfields method
- Add code in `facetoface_cm_info_view` to check if new config exists, create new property on sessionobject ([here](https://github.com/catalyst/moodle-mod_facetoface/blob/737840927bdf639b5593d079e6d1d3915ca8b40e/lib.php#L2887)), use the data from this new property to render the additional column
- Update CSS (eg update the f2foptions class to be a flexbox, row direction)

golenkovm:
Impleemented in https://github.com/catalyst/moodle-mod_facetoface/pull/235

Issue #199, opened 2025-05-05
Custom Field visibility


golenkovm:
Need to be able to select if a custom field is visible to students or not. This would make these fields available for reporting/export without student visibility. Similar set-up to the course custom fields visibility options.

There is a setting called `Show in exports and lists` that can be split into two.

golenkovm:
Implemented in https://github.com/catalyst/moodle-mod_facetoface/pull/230

PR #212, opened 2025-06-18
Issue #205:  Activity front screen changes

djarran:
> [!NOTE]
> Need to add additional functionality to hide column. Issue #199 will need to be merged first. 

**Description:** Show a selected custom field in the course page module view.

**Changes made**

1. Add admin setting to select which custom field is viewable on the course page
2. Display error when user attempts to delete a field this is selected for above setting

**Testing instructions:**

1. Create F2F module, create custom field
4. Check that this is viewable on the course page


customfield.php
```diff
@@ -42,6 +42,15 @@
     }
 }
 
+
+// If custom field is selected to be shown on custom page, alert user field cannot be deleted until
+// this is changed.
+$visiblefieldcolumn = get_config('facetoface', 'column');
+// $id = 0 occurs when creating a new customfield.
+if ($visiblefieldcolumn == $id && $id != 0) {
```
golenkovm: Simplify to `if (!empty($id) && $id == get_config('facetoface', 'column')) {` ?

customfield.php
```diff
@@ -42,6 +42,15 @@
     }
 }
 
+
```
golenkovm: Two empty lines in a row

settings.php
```diff
@@ -72,6 +72,19 @@
...
 
+$fields = $DB->get_records('facetoface_session_field', [], 'name', 'id, name');
+$options = array(0 => 'None');
+foreach ($fields as $field) {
+    $options[$field->id] = $field->name;
+}
+$settings->add(new admin_setting_configselect(
+    'facetoface/column',
+    'Visible field on course page',
+    'Show the field as an additional column on the session list view.',
```
golenkovm: Lang strings?

settings.php
```diff
@@ -72,6 +72,19 @@
     1
 ));
 
+$fields = $DB->get_records('facetoface_session_field', [], 'name', 'id, name');
+$options = array(0 => 'None');
+foreach ($fields as $field) {
+    $options[$field->id] = $field->name;
+}
+$settings->add(new admin_setting_configselect(
+    'facetoface/column',
```
golenkovm: It feels like `column` is not self explanatory enough name. May be change it to something like `displaycustomfield`?

lib.php
```diff
@@ -2897,6 +2897,24 @@ function facetoface_cm_info_view(cm_info $coursemodule) {
...
+                    // Get field title.
+                    $fieldname = $DB->get_field('facetoface_session_field', 'name', [
+                        'id' => $visiblefieldcolumn,
+                    ]);
+                    // Get field value for the session.
+                    $fieldvalue = $DB->get_field('facetoface_session_data', 'data', [
+                        'fieldid' => $visiblefieldcolumn,
+                        'sessionid' => $session->id
+                    ]);
+                    $sessionobject->field = (object)[
```
golenkovm: `$sessionobject->field` is a bit vague. Maybe make it at least `$sessionobject->customfield`?

styles.css
```diff
@@ -3,13 +3,14 @@
 }
 .f2fsession {
     width: 100%;
+    display: flex;
```
golenkovm, line 6: When setting a very long string for location the Options column gets unreasonably squeezed:

image

golenkovm (changes requested):
Thanks @djarran please find some feedback below

styles.css
djarran: This has been implemented like so:

image

djarran: @golenkovm: This is also an issue when on the module page.

image

Did we want to set a limit here too?

golenkovm: I think it's unlikely to have that long string with no spaces. How does it look when you have spaces in that long string?

djarran: Here's both pages with the same string you used:

image
image

golenkovm: Yep, looks good at the first glance :+1:

PR #213, opened 2025-06-18
Issue #199: Custom field visibility

djarran:
**Description:** Adds visibility setting for custom fields so that they can be only visible to everyone, teachers only, or nobody. 

Changes made:

1. use the same form element as course custom fields (Site Administration > Courses > Default settings > Course custom fields > Add a new custom field)
2. use the existing `editsessions` capability for the `Teachers` option, similar to how the visibility for course custom fields checks the `course/update` capability
3. The existing `showinsummary` option that was used for visibility in **BOTH** exports and lists is now only for exports
4. Accordingly, the language string for the `showinsummary` has been renamed to "Show in exports"
5. New `visibleto` column in `session_field` table

**Testing Instructions:**

1. Create test course with face-to-face module
6. Visit `admin/settings.php?section=modsettingfacetoface`
7. Create new custom field (i.e. Location)
8. Set `Visible to` setting to `Everyone`
9. Visit F2F module, create session, enter custom field value
10. Check that you can see the custom field in the module `view.php` page session list
11. Edit the custom field, set `Visible to` to `Teachers`
12. Login as student participant
13. Check that you can't see the custom field in the module `view.php` page session list


db/upgrade.php
```diff
@@ -859,5 +859,43 @@ function xmldb_facetoface_upgrade($oldversion=0) {
...
+        // Define field visibleto to be added to facetoface_session_field
+        $table = new xmldb_table('facetoface_session_field');
+        $field = new xmldb_field(
+            'visibleto',
+            XMLDB_TYPE_INTEGER,
+            '1',
+            null,
+            XMLDB_NOTNULL,
+            null,
+            '2', // Set default to visible for all.
```
golenkovm: Can we set it to `MDL_F2F_FIELD_VISIBLETOALL` instead of `2`?

db/upgrade.php
```diff
@@ -859,5 +859,43 @@ function xmldb_facetoface_upgrade($oldversion=0) {
...
+
+        // Conditionally launch add field visibleto.
+        if (!$dbman->field_exists($table, $field)) {
+            $dbman->add_field($table, $field);
+
+            // Get all existing records and set visibleto based on showinsummary value.
+            $records = $DB->get_records('facetoface_session_field');
+            foreach ($records as $record) {
+                $visibleto = 2;
+                if (isset($record->showinsummary) && $record->showinsummary == 0) {
```
golenkovm: In which case `$record->showinsummary` could be not set? It feels like this could be simplified a bit, eg `$visibleto = empty($record->showinsummary) ? MDL_F2F_FIELD_NOTVISIBLE : MDL_F2F_FIELD_VISIBLETOALL;`

db/upgrade.php
```diff
@@ -859,5 +859,43 @@ function xmldb_facetoface_upgrade($oldversion=0) {
...
+            '2', // Set default to visible for all.
+            'showinsummary'
+        );
+
+        // Conditionally launch add field visibleto.
+        if (!$dbman->field_exists($table, $field)) {
+            $dbman->add_field($table, $field);
+
+            // Get all existing records and set visibleto based on showinsummary value.
+            $records = $DB->get_records('facetoface_session_field');
```
golenkovm: Even though I'm not anticipating a high number of records in that table, I would still suggest to use `$DB->get_recordset()`

db/upgrade.php
```diff
@@ -859,5 +859,43 @@ function xmldb_facetoface_upgrade($oldversion=0) {
...
+            foreach ($records as $record) {
+                $visibleto = 2;
+                if (isset($record->showinsummary) && $record->showinsummary == 0) {
+                    $visibleto = 0;
+                }
+
+                $DB->execute(
+                    "UPDATE {facetoface_session_field} SET visibleto = ? WHERE id = ?",
```
golenkovm: Please use $DB->update_record() instead of $DB->execute()

classes/customfield_form.php
```diff
@@ -74,6 +74,18 @@ public function definition() {
         $mform->addElement('checkbox', 'showinsummary', get_string('setting:showinsummary', 'facetoface'));
         $mform->setDefault('showinsummary', true);
 
+        // Field data visibility.
+        $visibilityoptions = [
+                MDL_F2F_FIELD_VISIBLETOALL => get_string('customfield_visibletoall', 'facetoface'),
+                MDL_F2F_FIELD_VISIBLETOTEACHERS => get_string('customfield_visibletoteachers', 'facetoface'),
+                MDL_F2F_FIELD_NOTVISIBLE => get_string('customfield_notvisible', 'facetoface')
```
golenkovm, line 81: Sorting and numbering is a bit counterintuitive. I'd expect the items to be numbered 0, 1, 2 while in fact they are 2, 1, 0 here. Maybe swap the ids for MDL_F2F_FIELD_VISIBLETOALL and MDL_F2F_FIELD_NOTVISIBLE? As a result, the default value for the db column would be 0 (rather than 2)

lib.php
```diff
@@ -4301,6 +4309,23 @@ function facetoface_get_all_user_name_fields($returnsql = false, $tableprefix =
...
+ * @param bool $editsessions id of the question to test edit permission
+ * @return bool true if the current user can view custom fields, false otherwise
+ */
+function facetoface_can_view_field(int $visibility, bool $editsessions): bool {
+    if ($visibility == MDL_F2F_FIELD_NOTVISIBLE) {
+        return false;
+    } else if ($visibility == MDL_F2F_FIELD_VISIBLETOTEACHERS) {
+        return $editsessions;
+    } else {
+        return true;
```
golenkovm: I'd suggest to make MDL_F2F_FIELD_NOTVISIBLE as the final `else` and returning `false`. Currently, `true` is returned not only for the valid cases (`MDL_F2F_FIELD_VISIBLETOALL` and `MDL_F2F_FIELD_VISIBLETOTEACHERS + hascapability`), but also for all other invalid cases (eg, if for some reason any other number is passed.

golenkovm (changes requested):
Thanks @djarran The change is looking good and works as expected. Just a few minor tweaks are required.

classes/customfield_form.php
djarran: As in, you'd expect the default value to be `0`?

The way this is implemented is similar to other parts in Moodle, i.e:

1. [Course custom fields](https://github.com/moodle/moodle/blob/de18edbf585956f06e41165c61d0db8bdcf09049/course/classes/customfield/course_handler.php#L205)
2. [Question bank custom fields](https://github.com/moodle/moodle/blob/de18edbf585956f06e41165c61d0db8bdcf09049/question/bank/customfields/classes/customfield/question_handler.php#L290)

golenkovm: All right, happy to leave it aligned to other parts in Moodle then :ok_hand:

golenkovm (approved):
Thanks @djarran approving this and sending to UAT for testing

Issue #201, opened 2025-05-05
Restrict ability for student to cancel within "n" hours of session


golenkovm:
Need to be able to restrict ability for student to cancel within "n" hours of session, where "n" is a configurable number of hours as a setting for teachers/admins to set. Outside of that they can cancel themselves.

golenkovm:
Implemented in https://github.com/catalyst/moodle-mod_facetoface/pull/221

PR #215, opened 2025-06-18
Issue #201: Restrict ability for student to cancel

djarran:
**Description:** Add time restriction so that students can only cancel prior to some customisable period.

**Changes made:**

1. Add new admin setting `cancelrestriction`, `configduration` type with 'Enabled' flag option
2. Default value is two days
3. Disabled by default

**Testing Instructions:**

1. Create F2F module, create session in the future (i.e. 1 week)
2. Sign up to session
3. Check that you can still cancel
4. Visit plugin setting, enable time restriction (i.e. 2 weeks)
5. Check that you can no longer cancel the session

### DIFF
--- a/lang/en/facetoface.php
+++ b/lang/en/facetoface.php
@@ -159,6 +159,7 @@ $string['error:multipleusersmatched'] = 'Multiple users matched to identifier {$
 $string['error:addalreadysignedupattendee'] = '{$a} is already signed-up for this Face-to-Face activity.';
 $string['error:addattendee'] = 'Could not add {$a} to the session.';
 $string['error:cancellationsnotallowed'] = 'You are not allowed to cancel this sign-up.';
+$string['error:cancellationtooclose'] = 'You are not allowed to cancel this sign-up. Bookings can only be cancelled {$a} before session.';
 $string['error:cancelbooking'] = 'There was a problem cancelling your booking';
 $string['error:cannotemailmanager'] = 'Sent reminder mail for submission id {$a->submissionid} to user {$a->userid}, but could not send the message for the user\'s manager email address ({$a->manageremail}).';
 $string['error:cannotemailuser'] = 'Could not send out mail for submission id {$a->submissionid} to user {$a->userid} ({$a->useremail}).';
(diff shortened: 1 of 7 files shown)


lang/en/facetoface.php
```diff
@@ -410,6 +413,7 @@
 $string['setting:defaultcancellationmessage_caption'] = 'Cancellation message';
 $string['setting:defaultcancellationmessagedefault'] = 'This is to advise that your booking on the following course has been cancelled:
 
+
```
golenkovm: Looks like an empty line has been accidentally added

lang/en/facetoface.php
```diff
@@ -396,6 +397,8 @@
 $string['setting:addchangemanageremail_caption'] = 'Manager\'s email:';
 $string['setting:attendeesexportfields_caption'] = 'Attendees export fields';
 $string['setting:attendeesexportfields'] = 'Select the fields to be included in a session\'s exported list of attendees. This will be in addition to the attendee\'s first and last name.';
+$string['setting:cancelrestriction'] = 'Set the minimum number of hours before a session that students can cancel their booking. Within this time period, students cannot cancel themselves.';
```
golenkovm: Feels like this wording doesn't sound right. The setting doesn't set the _number of hours_. Instead, it could be days, weeks, etc. I'd suggest something like _Minimum time before a session that students can cancel their booking. Within this period, cancellations are not allowed_ or _Define the minimum time before a session that students are allowed to cancel their booking. Cancellations will not be allowed within this period_.

image

lang/en/facetoface.php
```diff
@@ -396,6 +397,8 @@
...
+$string['setting:cancelrestriction_caption'] = 'Cancellation Restriction';
```
golenkovm: Should it be `Cancellation restriction`?

golenkovm (approved):

golenkovm:
Hi @djarran 

The change looks good. I've added one commit to resolve the lang string issues.

The only thing I don't really like is that we throw an exception. I feel like it would be more user friendly if we didn't allow users to land on that page by greying out the cancellation links (`error:cancellationtooclose` lang string can go as a link title, that shows up when you hover the mouse). I found two of them:

1. On the `mod/facetoface/view.php?id=nnn` page:
image

2. On the `More info` page (`mod/facetoface/signup.php?s=4&backtoallsessions=1`):
image

Geying out example:
image

Could you please look into this and let me know if it's a quick change?

Cheers,
Misha

djarran:
Hi @golenkovm, thank you for updating the language strings. I have added functionality to disable the button if we are within the cancellation restriction.

golenkovm:
Thanks @djarran Approving this and sending to UAT for testing.

Issue #206, opened 2025-05-05
Visibility of sessions


golenkovm:
Need to be able to set up sessions, but choose to hide them. This is different to an expression of interest or waitlisted session where you want it visible but with no capacity. 

Implementation approach:
- Add new element to `session_form.php` for visibility setting
- Add new DB column `visible` in the `mdl_facetoface_sessions` table
- Add `visible` value from form submission to `$todb` in sessions.php [here](https://github.com/catalyst/moodle-mod_facetoface/blob/737840927bdf639b5593d079e6d1d3915ca8b40e/sessions.php#L191)
- Add check in [print_session_list_table](https://github.com/catalyst/moodle-mod_facetoface/blob/737840927bdf639b5593d079e6d1d3915ca8b40e/renderer.php#L74) for visibility; continue to next element in foreach if visibility = true; if teacher or admin roles, do not skip
- Add same check in [facetoface_cm_info_view](https://github.com/catalyst/moodle-mod_facetoface/blob/737840927bdf639b5593d079e6d1d3915ca8b40e/lib.php#L2842) function when submission (booking) doesn’t yet exist

PR #222, opened 2025-07-15
Issue #206: Add style changes when session is hidden

djarran:
**Description:** This adds the following:

1. When a session is hidden, its text in the session list will be dimmed
2. During an export, the status will be changed to show the visibility status if hidden


golenkovm:
Thanks @djarran please feel free to send it to UAT

PR #223, opened 2025-07-17
Issue #205: Activity front screen styling changes

djarran:
**Description:** Two changes:

1. Add fix so that if a field is selected to be shown on the course page for the module, it can be edited in the plugin settings
2. Apply CSS changes to custom field on course page view

CSS changes
image
image

Without field
image
image

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -3063,6 +3063,10 @@ function facetoface_get_visiblefield_data($session) {
         'sessionid' => $session->id
     ]);
 
+    if ($fieldname == null || $fieldvalue == null) {
+        return null;
+    }
+
     return (object)[
         'name' => $fieldname,
         'value' => $fieldvalue,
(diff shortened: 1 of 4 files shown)


lib.php
```diff
@@ -3042,6 +3038,30 @@ function facetoface_cm_info_view(cm_info $coursemodule) {
     $coursemodule->set_content($output);
 }
 
+function facetoface_get_visiblefield_data($session) {
```
golenkovm, line 3048: Could we please have phpdocs added to this new function along with expected param and return types. Related unit tests to cover this function would be also nice to have. Thanks

golenkovm (changes requested):
Please check the note under the `facetoface_get_visiblefield_data()` function.

lib.php
djarran: PHPDoc and unit tests have been added for this function now

golenkovm (approved):
Thanks @djarran Approving and merging.

Issue #225, opened 2025-07-21
Blocks cannot be added to some pages


djarran:
**Description:** Blocks cannot be added to the `attendees.php` and `editattendees.php` pages.

`view.php` (working)
<img width="954" height="348" alt="Image" src="https://github.com/user-attachments/assets/d9eee3f6-8da5-45b8-ab49-8fde5e012511" />

`attendees.php` (not working)
<img width="954" height="348" alt="Image" src="https://github.com/user-attachments/assets/b16e6249-b318-407e-a1b5-9f9e3e099f58" />

golenkovm:
Fixed in https://github.com/catalyst/moodle-mod_facetoface/pull/226

PR #226, opened 2025-07-21
Issue #225: Enable adding blocks on attendee pages

djarran:
**Description:** Resolves Issue #225. We remove the calls to `set_context` and `set_cm` because these are called from `require_course_login`, see following stack trace:

```
 moodle_page->set_cm pagelib.php:1254
 require_login moodlelib.php:2411
 require_course_login moodlelib.php:2766
 {main} attendees.php:78
```

```php
// require_course_login
    } else if (isloggedin() && !isguestuser()) {
        // User is already logged in. Make sure the login is complete (user is fully setup, policies agreed).
        require_login($courseorid, $autologinguest, $cm, $setwantsurltome, $preventredirect);

// require_login
        if ($cm) {
            $PAGE->set_cm($cm, $course);

// set_cm
        // Unfortunately the context setting is a mess.
        // Let's try to work around some common block problems and show some debug messages.
        if (empty($this->_context) or $this->_context->contextlevel != CONTEXT_BLOCK) {
            $context = context_module::instance($cm->id);
            $this->set_context($context);
        }
```

If we do not remove, we get the following error:

```
Whoops\Exception\ErrorException thrown with message "Coding problem: unsupported modification of PAGE->context from 70 to 50"

Stacktrace:
#1 debugging in /var/www/facetoface/lib/pagelib.php:1210
#0 moodle_page:set_context in /var/www/facetoface/mod/facetoface/attendees.php:186
```

Note that the `view.php` already has calls to `require_course_login` with this additional arguments:

https://github.com/catalyst/moodle-mod_facetoface/blob/MOODLE_403_STABLE/view.php#L78

### DIFF
--- a/attendees.php
+++ b/attendees.php
@@ -75,7 +75,7 @@ $cancellations = facetoface_get_cancellations($session->id);
  */
 $context = context_course::instance($course->id);
 $contextmodule = context_module::instance($cm->id);
-require_course_login($course);
+require_course_login($course, true, $cm);
 
 // Actions the user can perform.
 $canviewattendees = has_capability('mod/facetoface:viewattendees', $context);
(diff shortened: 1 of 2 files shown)


golenkovm:
Thanks @djarran 

Tested locally and the change looks good to me. Merging.

Cheers,
Misha

Issue #227, opened 2025-07-21
Roles with `moodle/course:view` are not enrolled when added as attendee


djarran:
**Description:** Users are enrolled into the course when added as attendees. However, this is not working for admin and manager roles as this call to `has_capability` returns true, and so we don't enter the conditional

https://github.com/catalyst/moodle-mod_facetoface/blob/b4887643901059c30acf6630706b9526a6215c05/editattendees.php#L94-L109

golenkovm:
Fixed in https://github.com/catalyst/moodle-mod_facetoface/pull/228

djarran:
Reopening this issue. We've previously addressed manually adding attendees to a session, but this is still occurring when users themselves (those with the system `moodle/course:view` capability) signup for a session.

golenkovm:
Implemented in https://github.com/catalyst/moodle-mod_facetoface/pull/237

PR #228, opened 2025-07-21
Issue #227: Enable managers and admins to be enrolled into course

djarran:
**Description:** Resolves Issue #227. Removes the redundant call to `has_capability`, and instead only uses the `is_enrolled` to check

### DIFF
--- a/editattendees.php
+++ b/editattendees.php
@@ -91,20 +91,17 @@ if (optional_param('add', false, PARAM_BOOL) && confirm_sesskey()) {
                 continue; // Invalid userid.
             }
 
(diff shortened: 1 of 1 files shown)


golenkovm:
Thanks @djarran 

I was wondering if that was done intentionally, but looking at the original code that comes from 2009 https://github.com/catalyst/moodle-mod_facetoface/blame/7700fccd58eb781703b869626c30dc796a998b13/editattendees.php#L95 makes me confident that it's not the case :smile:  

Tested the change and it works well. Merging.

Cheers,
Misha

PR #232, opened 2025-07-28
Issue #206: Add session status to 'Hidden' in session list view

djarran:
**Description:** If session is hidden, set the status to hidden on the session view list

Before:

image

After:

image


golenkovm:
Looks right, thank you @djarran

Issue #198, opened 2025-05-05
Marking attendance - The final session date to be stored as the course completion date


golenkovm:
In some cases there are multiple sessions that need to be attended to mark course as complete. If a teacher doesn't mark attendance on the date of the session, then the activity completion and course completion dates are incorrect. They aren't the date of attendance by the
student.

Course completion date needs to be recorded as the final session date in the course completion criteria. This can be configurable.

PR #234, opened 2025-07-30
Issue #198: Update course_completion_crit_compl table on module completion

djarran:
**Description:** There are two reports that show module completion:

1. Activity completion - `mdl_course_module_completion`
2. Course completion - `mdl_course_completion_crit_compl`

With the current code, we are only updating the `mdl_course_module_completion` entry. This PR adds support to also update the `mdl_course_completion_crit_compl` table


golenkovm:
Thanks @djarran The patch makes sense. Merging

PR #237, opened 2025-08-04
Issue #227: Fix enrolling during user signup

djarran:
**Description:** We've previously addressed manually adding attendees to a session (https://github.com/catalyst/moodle-mod_facetoface/pull/228), but this is still occurring when users themselves (those with the system moodle/course:view capability) signup for a session in a course they are not yet enrolled in.

This is because the user signup does not enrol users, whereas manually adding attendees does.

This PR aligns their behaviour

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -4420,6 +4420,22 @@ function facetoface_can_view_field(int $visibility, bool $editsessions): bool {
     }
 }
 
+function facetoface_enrol_user($context, $courseid, $userid) {
+    global $DB;
+
+    if (!is_enrolled($context, $userid)) {
+        $defaultroleid = null;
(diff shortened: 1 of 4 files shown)


lib.php
```diff
@@ -4420,6 +4420,22 @@ function facetoface_can_view_field(int $visibility, bool $editsessions): bool {
     }
 }
 
+function facetoface_enrol_user($context, $courseid, $userid) {
```
golenkovm: @djarran could you please add phpdoc to this function? Please add it as a new commit to facilitate the following review. Thanks

golenkovm:
Thanks @djarran LGTM :tm:

PR #244, opened 2025-08-18
Issue #243: Do not add calendar event to user updating settings

### LINKED ISSUE #243: Session calendar events created for user updating activity settings
**Description:** A calendar event is created for each session in an activity for the user that updates the module settings with the calendar option `Show entry on user's calendar` enabled. This is explicit behaviour defined in the code:

https://github.com/catalyst/moodle-mod_facetoface/blob/af4b43ea3accf10f093334658518b2704acb8b72/lib.php#L552-L558

This assumes that the user updating the activity settings is an attendee of the session, as per the `Show entry on user's calendar` help text:

> When active this setting adds a User Event entry to the calendar of an attendee of a Face-to-Face session. When turned off this prevents a duplicate event appearing in a session attendee's calendar, where you have calendar display settings set to Course or Site.

I don't think this is an assumption that should be made, as it's more likely for course creators or other users to create the session whom won't be attending the sessions themselves.

djarran:
**Description:** See issue #243 for more information

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -552,10 +552,6 @@ function facetoface_update_calendar_entries($session, $facetoface=null) {
     if ($facetoface->usercalentry) {
         // Get ALL enrolled/booked users.
         $users = facetoface_get_attendees($session->id);
-        // If session creator is not enrolled in the course, add the session to his/her events user calendar.
-        if (!in_array($USER->id, $users)) {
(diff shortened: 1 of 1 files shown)


golenkovm:
Thanks @djarran the change makes sense. In case it was really meant to create calendar even for those who can edit the activity but is not actually enrolled/signed up for the session it could be implemented as a configurable option. Will see if we ever get any feedback on this. Merging now.

PR #257, opened 2025-09-23
Reportbuilder fixes

djarran:
**Description:** 

1. Add Session ID column as this is typically the only way to identify a session besides time or description
2. Remove unneed JOINS on the details column of the session entity. Doesn't need to make reference to modules or course modules - it just needs to display the details column from `mdl_facetoface_sessions`


golenkovm:
Nice. Thank you @djarran Merged

PR #260, opened 2025-10-27
Send confirmation email

owenherbert-catalyst:
With the value set to false we do not get course booking notification emails sent through. Changing the value to `true` changes this default behavior and matches the default function parameter.

Other MR: https://github.com/catalyst/moodle-mod_facetoface/pull/261


golenkovm (approved):
Thanks @owenherbert-catalyst  looks pretty straightforward. Merging

PR #262, opened 2025-10-30
Fix schema alignment issues #259

### LINKED ISSUE #259: Schema alignment issue with facetoface_session_field visibleto
I'm testing an upgrade to 4.5 and the database schema alignment check is failing with:

```
facetoface_session_field

column 'visibleto' has default '2', expected '1' (I)
```
The install xml sets the default to 1, while the upgrade script uses MDL_F2F_FIELD_VISIBLETOALL, which is 2. 

https://github.com/catalyst/moodle-mod_facetoface/blob/e2e9e919149b9ae3ee2b4a8396bec66c4e8bd2ab/db/install.xml#L138

https://github.com/catalyst/moodle-mod_facetoface/blob/e2e9e919149b9ae3ee2b4a8396bec66c4e8bd2ab/db/upgrade.php#L865-L875

bwalkerl:
After speaking with @djarran the default should be 2. I updated install.xml and used XMLDB editor to generate the upgrade script (modified to include MDL_F2F_FIELD_VISIBLETOALL) for users who installed the plugin after the patch. 

Closes #259


bwalkerl:
The tests are failing because of newly added codechecker detections in moodle-plugin-ci 4.5.9. While these can mostly be fixed with phpcbf, there's a lot so it might be better left to someone with more experience with the plugin.

PR #266, opened 2025-11-07
[#265] Allow uploading bookings for all session times

matthewhilton:
Closes #265 

Given users can already do this one-by-one via the gui (there is no restriction there about booking in progress or past sessions), I think it makes sense to match this with the CSV upload. Therefore I didn't gate it behind a config.

Testing is covered by unit tests

PR #267, opened 2025-11-07
misc: fixup code standards

matthewhilton:
This is a big fixup PR to get code standards (and thereby CI) green again

Comprises of:
- 85% `phpcbf` auto-fixes
- 15% manually adding/fixing phpdocs like `@license` and removing double docs in class files or adding missing docs
- Added one-class-per-file rule ignore in `lib.php` as there are 3 classes in here, I'd rather not move them as that could cause issues if say other plugins depend on these 
- Small test fixup/regression from https://github.com/catalyst/moodle-mod_facetoface/pull/266 - I initially thought the tests passed but these were actually the _behat_ tests not the _phpunit_ test, the phpunit tests had this failure in the previous PR but was masked due to the code standards issue

I also disabled phpcpd, as it was erroring BUT there is a warning about `The phpcpd command is deprecated and will be removed in 5.0.0. No replacement is planned.`, so I figured it's not worth fixing.


backup/moodle1/lib.php
```diff
@@ -19,20 +19,20 @@
  * Copyright (C) 2011-2013 Totara LMS (http://www.totaralms.com)
  * Copyright (C) 2014 onwards Catalyst IT (http://www.catalyst-eu.net)
  *
- * @package    mod
+ * @package    mod_facetoface
  * @subpackage facetoface
```
Peterburnett: We should probably remove these subpackage references in general

classes/event/add_session.php
```diff
@@ -14,17 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
-/**
- * Copyright (C) 2007-2011 Catalyst IT (http://www.catalyst.net.nz)
```
Peterburnett, line 18: We should probably maintain these copyright statements in the class comment

classes/event/add_session_failed.php
```diff
@@ -14,17 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
-/**
- * Copyright (C) 2007-2011 Catalyst IT (http://www.catalyst.net.nz)
```
Peterburnett, line 18: Same, maintain

tests/reportbuilder/datasource/facetoface_test.php
```diff
@@ -1,4 +1,18 @@
 <?php
+// This file is part of Moodle - https://moodle.org/
```
Peterburnett: This appears to be a duplicate

Peterburnett (changes requested):
Couple of minor mistakes noted, LGTM once fixed

backup/moodle1/lib.php
matthewhilton: yep agreed, removed

Peterburnett (approved):
LGTM

PR #271, opened 2025-11-11
Upload changes

### LINKED ISSUE #269: CSV upload of bookings without optional columns fails
Lang string for discountcode and calendar type mention they are optional, but if you remove them you get an error

```
Mismatched number of fields in the uploaded file on row {$a}.

[More information about this error](https://docs.moodle.org/405/en/error/mod_facetoface/error:bookingsuploadfileheaderfieldmismatch)
Debug info:
Error code: error:bookingsuploadfileheaderfieldmismatch
Stack trace:

    line 150 of /mod/facetoface/classes/booking_manager.php: core\exception\moodle_exception thrown
    line 177 of /mod/facetoface/classes/booking_manager.php: call to mod_facetoface\booking_manager->get_iterator()
    line 74 of /mod/facetoface/upload.php: call to mod_facetoface\booking_manager->validate()
```

### LINKED ISSUE #270: Allow csv upload directly into attendance status
E.g. to backfill attendance

This will need to sign them up first, and then take attendance, and then handle any edge cases e.g. future sessions.

matthewhilton:
2 somewhat related CSV upload changes:
Closes #269 
Closes #270 

**Changes**
1. Improve docs and handling of optional headers. I believe before they were optional in the sense that the headers must be there, but the values could be empty. Now, you don't need to include the column at all.
2. Be able to book directly into an attendance status.
    1. Updated lang string to explain all allowed statuses (we don't handle some e.g. `cancelled`)
    2. Added some notes/caveats to help lang string
    3. Updated tests to match

**Testing**
This is covered by new unit tests


classes/booking_manager.php
```diff
@@ -403,9 +454,15 @@ public function process() {
...
+                $attendancestatuses = [MDL_F2F_STATUS_NO_SHOW, MDL_F2F_STATUS_PARTIALLY_ATTENDED, MDL_F2F_STATUS_FULLY_ATTENDED];
+
+                $issignup = in_array($statuscode, $signupstatuses);
+                $isattendance = in_array($statuscode, $attendancestatuses);
+
                 // Handle signups.
-                if (in_array($statuscode, [MDL_F2F_STATUS_BOOKED, MDL_F2F_STATUS_WAITLISTED])) {
-                    if ($statuscode === MDL_F2F_STATUS_BOOKED && !$session->datetimeknown) {
+                if ($issignup) {
+                    if (($statuscode === MDL_F2F_STATUS_BOOKED) && !$session->datetimeknown) {
```
keevan:
```suggestion
                    if ($statuscode === MDL_F2F_STATUS_BOOKED && !$session->datetimeknown) {
```
Unnecessary grouping

classes/booking_manager.php
```diff
@@ -414,25 +471,37 @@ public function process() {
...
-                    ])
-                ) {
+                if ($isattendance) {
+                    // If booking into attendance but the user hasn't been signed up yet (e.g. directly marking as attended),
+                    // then sign the user up.
+                    $alreadysignedup = $DB->record_exists('facetoface_signups', ['sessionid' => $session->id, 'userid' => $user->id]);
+                    if (!$alreadysignedup) {
+                        // We use booked/waitlisted for making the signup,
+                        // and then the actual status from the CSV when taking attendance.
+                        $signupstatus = $session->datetimeknown ? MDL_F2F_STATUS_BOOKED : MDL_F2F_STATUS_WAITLISTED;
```
keevan, line 490: I'm guessing if the datetime is known, waitlisted is discarded as a valid default option, because we're filling the entry to ensure we can take attendance further down (i.e. it's not important what the value is)

classes/form/upload_bookings_form.php
```diff
@@ -59,11 +60,14 @@ public function definition() {
         $mform->setType('csvfile', PARAM_INT);
         $mform->addRule('csvfile', get_string('required'), 'required', null, 'client');
 
+        // Allowed statuses minus the '' one, as we have additional help text to explain '' status specifically.
+        $allowedstatuses = array_filter(booking_manager::get_allowed_session_statuses(), fn($status) => !empty($status));
```
keevan:
```suggestion
        $allowedstatuses = array_filter(booking_manager::get_allowed_session_statuses());
```
By default it strips out empty/falsey entries, so you shouldn't need to supply any custom function to handle it.

classes/form/upload_bookings_form.php
```diff
@@ -59,11 +60,14 @@ public function definition() {
...
+        $help = nl2br(get_string('facetoface:uploadbookingsfiledesc', 'mod_facetoface', implode(', ', $allowedstatuses)));
```
keevan, line 65: No action here really, just wanted to mark that php code can be so ugly sometimes.

Maybe it's slightly easier on the eyes if one splits the burden. (assignment, then formatting/print)
```php
$help = get_string('this-really-long-thing' ... );
...
$mform->addElement(
  // ...
  '',
  nl2br($help),
);
```
:shrug:

tests/upload_test.php
```diff
@@ -957,4 +958,209 @@ public function test_upload_attendance_for_all_session_statuses(int $timestart,
...
+     *
+     * @param string $contents csv contents
+     * @param array $sessiondates dates for session, each entry should be in the form ['timestart' => xxx, 'timefinish' => xxx].
+     * @return array array of [$bookingmanager, $session, $student]
+     */
+    private function setup_csv_test(string $contents, array $sessiondates = []): array {
+        global $USER;
+        $this->setAdminUser();
+
+        // This is what facetoface does internally when you don't set a time.
```
keevan, line 1016: > // This is what facetoface does internally when you don't set a time.

A reference might be useful.

tests/upload_test.php
```diff
@@ -957,4 +958,209 @@ public function test_upload_attendance_for_all_session_statuses(int $timestart,
...
+    /**
+     * Provides CSV processing test cases
+     * @return array
+     */
+    public static function csv_process_provider(): array {
+        // Build a matrix of status, date (past present future and unknown).
+        $dateoptions = [
+            'past' => ['timestart' => time() - DAYSECS * 2, 'timefinish' => time() - DAYSECS],
+            'present' => ['timestart' => time() - DAYSECS, 'timefinish' => time() + DAYSECS],
+            'future' => ['timestart' => time() + DAYSECS, 'timefinish' => time() - DAYSECS * 2],
```
keevan:
```suggestion
            'future' => ['timestart' => time() + DAYSECS, 'timefinish' => time() + DAYSECS * 2],
```
Should timefinish be after timestart?

tests/upload_test.php
```diff
@@ -957,4 +958,209 @@ public function test_upload_attendance_for_all_session_statuses(int $timestart,
...
+                }
+
+                // If no status is given, it should be either waitlisted or booked
+                // depending on the time of the session.
+                $expectedstatus = $status;
+                if ($status == '' && $datelabel == 'unknown') {
+                    $expectedstatus = 'waitlisted';
+                } else if ($status == '' && $datelabel !== 'unknown') {
+                    $expectedstatus = 'booked';
+                }
```
keevan: I think generally speaking in php-land, it's preferable to use `===` instead of `==` where possible. The exception would be if type coercion is desired. 

Alternatively using a method like `is_null` or in this case, `empty(...)` would work too.

tests/upload_test.php
```diff
@@ -957,4 +958,209 @@ public function test_upload_attendance_for_all_session_statuses(int $timestart,
...
+
+        if ($issignup) {
+            $this->assertEquals($expectedstatus, $status);
+        }
+
+        if ($isattendance) {
+            $this->assertEquals('booked', $status);
+            $attendance = facetoface_get_attendee($session->id, $user->id);
+            $this->assertEquals($expectedstatus, facetoface_get_status($attendance->statuscode));
+        }
```
keevan: Perhaps it might be easier to digest if `$status` was rewritten to `$storedstatus`.

Also, is the `$expectedstatus` coming from the CSV, or is this the expected end state (status) for the attendance entry? If only showing end state, then does `$signupstatuses` need the `''` (empty) string entry? Or maybe I'm missing something and there is a chance it is validly stored as an empty value somehow?

classes/booking_manager.php
matthewhilton: Fair point - It probably doesn't matter, but on the off chance it does (given the f2f logic is pretty spaghetti) i'd rather just simulate what it does when you add/remove someone via the UI i.e. using waitlisted if time is not known, otherwise booked

tests/upload_test.php
matthewhilton: In csv_process_provider it works out what the actual status should be if you were to enter `''` (empty)

https://github.com/catalyst/moodle-mod_facetoface/blob/672a4cd7d9f70642abf1e3cc9dfdb1667b08f5fa/tests/upload_test.php#L1105-L1108

But I think I forgot to remove the `''` from the `$signupstatuses` array, so that is likely the source of confusion

keevan:
Looks good to me. Merging.

PR #277, opened 2025-12-15
Add idnumber to attendees table

owenherbert-catalyst:
This merge requests introduces a new idnumber column for the attendees table. This new column can be toggled on/off with a new configuration setting. This column will also abide by the Moodle userpolicies configuration. The attendees download file will include the new field when enabled.

image

**What's been done?**
- Add new `idnumber` column to page and download file.
- New `attendeesshowidnumbers` config setting.

**Testing instructions:**
1. Install this plugin branch on a Moodle 4.5 site.
2. Create an admin user and a student.
3. Confirm new "Show id number in attendees table" plugin setting exists. Set it to true.
4. Add idnumber to "Show user identity" Moodle setting. ({domain}/admin/settings.php?section=userpolicies)
5. Create a test course.
6. Add F2F to the test course.
7. Enroll all users to the test course.
8. Create a F2F session.
9. Add all users to the F2F session.
10. Confirm all users appear in the attendees table and the columns are populated with data. Also check downloaded file.
11. Run unit tests with vendor/bin/phpunit --testsuite mod_facetoface_testsuite.


attendees.php
```diff
@@ -257,13 +257,20 @@
...
         $table->align = ['left'];
-        $table->size = ['70%'];
+        $table->size = ['50%'];
 
         $showusername = facetoface_should_attendees_show_usernames();
         if ($showusername) {
             $table->head[] = get_string('username');
             $table->align[] = 'left';
-            $table->size[] = '30%';
+            $table->size[] = '25%';
```
Peterburnett: This size feels like it should respect which columns are active.

owenherbert-catalyst: I agree, now that we have a few columns this value should be calculated dynamically.

owenherbert-catalyst: I've set the size to be `auto` so that width is calculated dynamically.

Peterburnett: Great this is looking much better

PR #281, opened 2026-03-26
Addressed bug where person who did not attend session completes course.

abhinavgandham:
## Issue Summary
The course completion had an issue with face-to-face where if a student could be marked as course complete, even if they did not attend a session.

### Detailed Scenario
1. A session is created and a student books in for that session.
2. The student does not attend the session, and is marked as "Not attended".
3. The student books in for another session.
4. The "Calculate regular completion" task gets run.
5. Once the task finishes, the student is marked as "Course complete" in the course completions report, with the time of completion set to the end time of that other booked in session.

This PR addresses this bug by making sure that the course completion is set to complete only if the student actually fully attended the session.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2930,25 +2930,25 @@ function facetoface_take_individual_attendance($submissionid, $grading) {
     // Reset completion if set up.
     if ($record->completionattendance) {
         $course = $DB->get_record('course', ['id' => $record->course], '*', MUST_EXIST);
-        $completion = new completion_info($course);
+        $completion = new \completion_info($course);
         $cm = get_coursemodule_from_instance('facetoface', $record->id, $course->id);
         if ($completion->is_enabled($cm)) {
             // Update/create completion data
             $completion->update_state($cm, COMPLETION_UNKNOWN, $record->userid, false);
-
-            if ($record->datetimeknown && get_config('facetoface', 'sessioncompletiondate')) {
-                // Get existing completion data, modify state, save, and update completion.
-                $data = $completion->get_data($cm, false, $record->userid);
-                $data->timemodified = $record->timefinish;
-                $completion->internal_set_data($cm, $data);
-                $completion->update_state($cm, COMPLETION_UNKNOWN, $record->userid, false);
-
-                // Update the course completion criteria entry.
-                $criteras = $completion->get_criteria(4); // 4 = completion_criteria_activity
-                foreach ($criteras as $criteria) {
-                    if ($criteria->module == 'facetoface' && $criteria->moduleinstance == $cm->id) {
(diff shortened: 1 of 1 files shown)


lib.php
```diff
@@ -2930,25 +2930,37 @@ function facetoface_take_individual_attendance($submissionid, $grading) {
...
-                // Update the course completion criteria entry.
-                $criteras = $completion->get_criteria(4); // 4 = completion_criteria_activity
-                foreach ($criteras as $criteria) {
-                    if ($criteria->module == 'facetoface' && $criteria->moduleinstance == $cm->id) {
-                        $criteriacompletion = $completion->get_user_completion($record->userid, $criteria);
-                        $criteriacompletion->mark_complete($record->timefinish);
+
+                // Check if this face-to-face activity is in completion criteria
+                $criteria = $completion->get_criteria(4); // 4 = completion_criteria_activity
+                $is_completion_criteria = false;
```
djarran: It's more readable with underscores, but Moodle standard state that variables shouldn't contain underscores :/

https://moodledev.io/general/development/policies/codingstyle#variables

lib.php
```diff
@@ -2930,25 +2930,37 @@ function facetoface_take_individual_attendance($submissionid, $grading) {
...
+                foreach ($criteria as $criterion) {
+                    if ($criterion->module == 'facetoface' && $criterion->moduleinstance == $cm->id) {
+                        $is_completion_criteria = true;
+                        break;
+                    }
+                }
+                
+                // Set the course completion date if the student fully attended the session.
+                if ($is_completion_criteria && $grading == MDL_F2F_STATUS_FULLY_ATTENDED) {
```
djarran: Since we only want to run this logic if `$grading == MDL_F2F_STATUS_FULLY_ATTENDED`, are we able to move this check earlier in the condtional, or do an early return? 

In this way, we'd only need to do the following once:

```php
$criteria = $completion->get_criteria(4); // 4 = completion_criteria_activity
foreach ($criteria as $criterion) {
    if ($criterion->module == 'facetoface' && $criterion->moduleinstance == $cm->id) {
```

djarran (changes requested):
Thanks @abhinavgandham, this resolves this issue. I followed the scenario you gave and was able to replicate the issue without this patch. After applying the patch, the issue no longer occurred.

Just a change and a question and we can get this one merged.

lib.php
djarran: We may also need to consider the level of attendance required set in the activity

image

I believe this is set in the `completionattendance` column in the `facetoface` table

djarran:
Thanks @abhinavgandham, this is looking good now. Merging

PR #282, opened 2026-04-09
Add new sessiondateshort placeholder

djarran:
**Description:** The `sessiondate` placeholder also includes the time:

```
10 April 2026, 11:40 AM - 11 April 2026, 11:40 AM
```

Instead of making a breaking change, we add a new optional placeholder to display the date like so:

```
Multi-day session: 15/04/26 - 16/04/26

Single-day session: 16/04/26
```

The exact string output by this placeholder depends on the environment's `strftimedatefullshort` language string. By default in Moodle, this is:

```
$string['strftimedatefullshort'] = '%d/%m/%y';
```

### Test instructions

1. Update the plugin with this patch (need to purge for new language string)
2. Create a F2F activity
3. Go to Activity > Settings > Confirmation message
4. Edit the subject and replace `sessiondate` with `sessiondateshort`
5. Create a new session with a date in the future
6. Add an attendee
7. Check emails
8. Confirm that the session date uses the new short format

### DIFF
--- a/classes/session.php
+++ b/classes/session.php
@@ -32,20 +32,26 @@ class session {
      * Get a human-readable string of the dates for a session instance.
      *
      * @param \stdClass $sessiondate Object containing a start and finish time for a session.
+     * @param string $dateformat Lang string key for the date format.
      * @return string Date string. If session is only over a day, it just returns one date, otherwise it will show a range.
      *
      * @throws \moodle_exception
      */
-    public static function get_readable_session_date(\stdClass $sessiondate): string {
+    public static function get_readable_session_date(\stdClass $sessiondate, string $dateformat = 'strftimedate'): string {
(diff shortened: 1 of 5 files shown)


abhinavgandham:
### Testing Results

**Scenario 1: Single day session**

With `sessiondate` placeholder:

image

With `sessiondateshort` placeholder:

image

**Scenario 2: Multi-day session**

With `sessiondate` placeholder:

image

With `sessiondateshort` placeholder:

image


### Conclusion
The patch seems to be working as expected, the `sessiondateshort` placeholder is properly simplifying the format.

abhinavgandham:
Code looks good to me. Some warnings with codechecker and missing param section for phpdocs, including the shortdate test, but I think we can leave it as is for now, and probably plan a plugin cleanup for all this in a future time. All PHPUnit tests are passing which is good.

abhinavgandham (approved):
LGTM

PR #284, opened 2026-07-21
UI improvements

rhell4:
Changes:
---
- Improves the sessions table appearance
- Adds a new page for users to view all of their sessions across the site

Sessions table:
---
| Before | After |
| ---------- | --------|
| image  | image |

New page:
---
image


classes/my_bookings_table.php
```diff
@@ -0,0 +1,220 @@
...
+     * Sets up the table.
+     *
+     * @param \core\url $url The base URL.
+     * @param ?string $sessiontype The session type to display (upcoming or previous).
+     */
+    public function __construct(\core\url $url, ?string $sessiontype = null) {
+        global $DB, $USER;
+
+        parent::__construct('facetoface_mybookings');
+        $this->context = \core\context\system::instance();
```
Peterburnett: Should this be at the system context? My initial naive thought would be that this better aligns with the user context

classes/my_bookings_table.php
```diff
@@ -0,0 +1,220 @@
...
+            'sessiontime' => get_string('time', 'facetoface'),
+            'status' => get_string('status', 'facetoface'),
+        ];
+
+        $this->define_columns(array_keys($columns));
+        $this->define_headers(array_values($columns));
+        $this->define_baseurl($url);
+        $this->no_sorting('status');
+
+        $uniqueid = $DB->sql_concat(...['s.id', '\'-\'', 'sd.id']);
```
Peterburnett: why spread here instead of just calling sql_concat('s.id', ....);

classes/my_bookings_table.php
```diff
@@ -0,0 +1,220 @@
...
+                   f.name AS facetoface,
+                   sd.timestart AS sessiontime,
+                   sd.timestart,
+                   sd.timefinish,
+                   ss.statuscode';
+        $from = 'FROM {facetoface_signups} su
+                 JOIN {facetoface_signups_status} ss ON ss.signupid = su.id AND ss.superceded = 0
+                 JOIN {facetoface_sessions} s ON s.id = su.sessionid
+                 JOIN {facetoface} f ON f.id = s.facetoface
+                 JOIN {course} c ON c.id = f.course
```
Peterburnett, line 102: It feels like we should add some visibility restrictions here. If a course/activity is hidden, I don't think we should render sessions for it. I know it adds a bit of pain to the data loader (don't think doing it in sql is the easiest way), but it will prevent dead links from being rendered for example (eg you click to go to activity, blocked by visibility restrictions).

classes/my_bookings_table.php
```diff
@@ -0,0 +1,220 @@
...
+        $sql = "SELECT {$this->sql->fields}
+                {$this->sql->from}
+                {$this->sql->where}
+                {$order}";
+
+        if (!$this->is_downloading()) {
+            $this->rawdata = $DB->get_records_sql($sql, $this->sql->params, $this->get_page_start(), $this->get_page_size());
+        } else {
+            $this->rawdata = $DB->get_records_sql($sql, $this->sql->params);
+        }
```
Peterburnett, line 201: Ah here is probably the best spot to do the visibility checks on course/activity

Peterburnett: probably need to check the f2f view permission on the activity context as well.

classes/my_bookings_table.php
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Display course name column as link
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_coursename(\stdClass $record): string {
+        return html_writer::link(
+            new \moodle_url('/course/view.php', ['id' => $record->courseid]),
```
Peterburnett, line 214: Should these be core\urls instead? not sure on the support matrix for those vs this branch, just flagging it

my_bookings.php
```diff
@@ -0,0 +1,71 @@
...
+require_once(__DIR__ . '/../../config.php');
+
+use mod_facetoface\my_bookings_table;
+
+require_login(null, false);
+
+$sessiontype = optional_param('sessiontype', my_bookings_table::UPCOMING, core\param::TEXT->value);
+
+$url = new core\url('/mod/facetoface/my_bookings.php', ['sessiontype' => $sessiontype]);
+$context = core\context\system::instance();
```
Peterburnett: Again this feels like it should be user context

renderer.php
```diff
@@ -157,91 +169,138 @@ public function print_session_list_table(
...
+                    )
+                );
+                $divider = new filler();
+                $divider->primary = false;
+                $actionsmenu->add($divider);
+                $actionsmenu->add(
+                    new link_secondary(
+                        new core\url('attendees.php', ['s' => $session->id, 'download' => 'xlsx']),
+                        new pix_icon('f/spreadsheet', get_string('downloadexcel')),
+                        get_string('downloadexcel')
```
Peterburnett: Can we use a different string to mention its the attendance data specifically. Currently its not super clear what data you are downloading in the context menu

classes/my_bookings_table.php
rhell4: I propose we still include the text details but check for visibility if it should be a link or just plain text, since to book a session they would have had visibility at some point, so I don't think we necessarily need to hide it altogether. This will make it easier to manage the visibility without creating pages with missing rows or doing it via the SQL. Thoughts?

classes/my_bookings_table.php
rhell4: Ah this was my mistake it should have been `core\url`, but since you bring up support of older versions the separating of classes into namespaces was not done in 4.3, so I will need to make changes so everything is 4.3 compatible.

classes/my_bookings_table.php
rhell4: Ignore the above, we decided to just remove the rows entirely.

classes/my_bookings_table.php
rhell4: Yeah you're right there's no need to spread here, was leftover after I made some tweaks, fixed.

classes/my_bookings_table.php
rhell4: You're right user context makes more sense, especially since I moved it to live under a user's profile.

classes/my_bookings_table.php
rhell4: I have adjusted how the data is retrieved so it now filters out any f2f activities the user should not be able to see. Unfortunately due to pagination/total counts we need to fetch all of the sessions and then check their visibility to get the correct total and number of pages, but the alternative is to do it via SQL and I think that would actually perform worse and create too much complexity with roles/perms/etc..

I also double checked and the `uservisible` variable from `modinfo` does include the f2f view perms.

Let me know what you think.

classes/my_bookings_table.php
rhell4: I've gone back and changed everything so it is 4.3 compatible, which should be fine as it's still supported in the latest version.

renderer.php
rhell4: Adjusted these to "Download attendees in `{format}` format".

classes/my_bookings_table.php
owenherbert-catalyst, line 1: "Signup" label shown to already-booked users. 

Image

classes/my_bookings_table.php
```diff
@@ -0,0 +1,272 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+declare(strict_types=1);
+
+namespace mod_facetoface;
+
+use core_reportbuilder\local\aggregation\count;
```
owenherbert-catalyst: Is this import used?

my_bookings.php
```diff
@@ -0,0 +1,71 @@
...
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require_once(__DIR__ . '/../../config.php');
+
+use mod_facetoface\my_bookings_table;
+
+require_login(null, false);
+
+$sessiontype = optional_param('sessiontype', my_bookings_table::UPCOMING, PARAM_TEXT);
```
owenherbert-catalyst, line 33: This can cause a fatal error if I navigate to, for example: `/mod/facetoface/my_bookings.php?sessiontype=invalid` . Can this be handled differently?

classes/my_bookings_table.php
```diff
@@ -0,0 +1,272 @@
...
+
+    /**
+     * Display status column
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_status(\stdClass $record): string {
+        if ($record->datetimeknown) {
+            $record->sessiondates = facetoface_get_session_dates($record->sessionid);
```
owenherbert-catalyst: Is this dead code? Do we use `$record->sessiondates` ?

lib.php
```diff
@@ -4805,6 +4805,27 @@ function facetoface_should_attendees_show_idnumbers(): bool {
...
+/**
+ * Add nodes to myprofile page.
+ *
+ * @param \core_user\output\myprofile\tree $tree Tree object
+ * @param stdClass $user user object
+ * @param bool $iscurrentuser
+ * @param stdClass $course Course object
+ * @return void
+ */
+function mod_facetoface_myprofile_navigation(core_user\output\myprofile\tree $tree, $user, $iscurrentuser, $course) {
```
owenherbert-catalyst, line 4817: If I click on 'My bookings' on a users profile page it will show my (admin user) bookings instead of the user.

Image

owenherbert-catalyst (commented):
Hey @rhell4 , just a few comments. Unit tests and code checker are looking good!

classes/my_bookings_table.php
rhell4: The button was correct but the label was incorrect, changed it to "More info".

classes/my_bookings_table.php
rhell4: No that was a leftover after some changes, have now removed it.

classes/my_bookings_table.php
rhell4: Yes this is dead code, before I was going to make it like the regular facetoface session table that shows all the session dates in the one row, I scrapped that idea and separated it into one row per session date. Removed this left over code.

lib.php
rhell4: The original idea was this was only to be used by the users to see their own bookings, but it does make sense and would serve a purpose to be able to see someone else's bookings. I've added a new capability and swapped everything over so a user id can be passed to see a different user's bookings.

my_bookings.php
rhell4: Good catch, I've made some changes so that if an incorrect session type is passed in the URL the user is redirected to the default page (which is upcoming session).

PR #288, opened 2026-08-28
Fix attendance completion session date

NilakshaPerera:
Fixes the `sessioncompletiondate` handling when attendance completion is updated.

The completion timestamp now uses the session finish time consistently, including when:
- the course does not have activity completion criteria configured
- the user is marked partially attended
- the session has multiple date records

## Details

Previously, the completion date logic only updated the course completion criterion path, so activity completion could still be stamped with the attendance marking time.

The old check also compared the attendance grade against the required completion status. This broke partial attendance because partial attendance uses status `90` but grade `50`.

This change:
- uses the current signup status code when deciding whether attendance completed the activity
- uses the latest session finish time for multi-date sessions
- updates both activity completion and matching course completion criteria
- skips backdating when the session has no known finish date or the activity is not complete

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -2916,16 +2916,27 @@ function facetoface_approve_requests($data) {
 function facetoface_take_individual_attendance($submissionid, $grading) {
     global $USER, $CFG, $DB;
 
+    require_once($CFG->dirroot . '/completion/criteria/completion_criteria.php');
+
     $timenow = time();
+
+    /*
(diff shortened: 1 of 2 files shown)


danmarsden:
Thanks @NilakshaPerera - it may take some time before we get to reviewing this, but in the meantime can you please confirm which Moodle versions you have tested this with on your end?

thanks,

